PR #6, opened 2021-05-27
Sentry SDK and Database

thinkocapo:
## Done
- **Release Management** for test and prod environments is complete. This includes setting a Release version according to the calendar, regardless of what npm command or test vs prod environment was used. Easy to see in the code that you can override this. **This is essential to Release Health properly functioning, as well as being able to see errors over time per Release in Discover**, and tracing frontend to backend errors and making sure they're on set releases.

**React's** package.json
```
 "start": "REACT_APP_RELEASE=`./release.sh` react-scripts start",   
 "build": "REACT_APP_RELEASE=`./release.sh` react-scripts build",
```
**Flask's** run.sh
```
RELEASE=`../release.sh`
```
**Deployment's** deploy.sh
```
RELEASE=`./release.sh`
```
^ so that your source maps get the same release as whatever you're deploying, as well as running locally.
- the **backend url** now comes from .env if `npm run build` is run, or else it's hard coded as localhost. This eliminates the reliance on a third variable like backend_local_url as was used in sentry-demos/tracing.
- read from **cloudsql db**
- `deploy.sh` script simplified over makefile
- flask's `run.sh` is same as our former 'make deploy', but now it sets the Release for you too.
- removed React.StrictMode because that was interfering with how the Sentry Profiler and a transaction's spans were getting captured, per convo with Abhijeet.
- started a Features/Integrations list in the Readme so you can identify sooner if the Profiler or React Router and other integrations are used, rather than have to go hunting through the code. This is a Flagship App so this needs to be clear.
- added SqlAlchemy Integration, but it turns out SELECT * FROM tools was appearing as a span anyways without it.
- decided will use a mix of both React Class Components and React Function Components, per discussions with Abhi and findings in our sentry-demos/react-native demo.

## Next
- additional database tables, for well performing and poor performing queries, to show the before/after state.
- make BACKEND URL global as multiple routes will be using it.
- **fix** the source maps processing warning. It deminifies fine, but there's a warning about invalid column, may be due to CreateReactApp's chunking.

## Questions
- do even need this anymore `or release()` as the .env var is getting the same variable, could cause confusion. let users define there own if they want, and not compete (i.e. have to comprehend) with 2 different options in place.
- Anything basic to SqlAlchemy config we can add, so that SqlAlchem middleware operations get captured as spans?
- Anything basic to Flask configuration we can add, so that Flask middleware operations get captured as spans?
#### backend url
```
console.log("PROCESS.ENV.TEST", process.env.TEST)
```
```
could get host from window.location.host, but would have to swap "flask" into there...BAD
  BACKEND = process.env.REACT_APP_BACKEND
```

```
// did not have globals
console.log("this.context", this)
```

```
//src/index.js
const THEBACKEND = "thebackend"
const THEBACKEND = "thebackend"
export default THEBACKEND
export { THEBACKEND }
```


src/index.js
```diff
@@ -31,27 +31,31 @@ import productFour from './components/products/4';
 
 import plantsBackground from './assets/plants-background-img.jpg';
 
-const tracingOrigins = ['localhost', 'empowerplant.io', /^\//];
+const tracingOrigins = ['localhost', 'empowerplant.io', 'run.app', 'appspot.com', /^\//];
```
thinkocapo, line 34: for App Engine, as well as Cloud Run, if that ever gets used.

PR #15, opened 2021-06-23
adding checkout test

ndmanvar:
https://app.saucelabs.com/tests/b2d5d2a5e4444c9fa326fdeef581b410
^ succesful test

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,72 @@
+# Test Data Automation
+Runs automted tests against Sentry demos on GCP, in order to generate errors and transactions to be sent to Sentry.io
+
+## Components / Moving parts
+- `create_job.sh` -> creates GCP cron job which hits Travis requests APIs to trigger build
+- `.travis.yml` -> runs automated tests / simulations
+- `conftest.py` -> Sauce Labs configuration (browsers) for frontend_tests
+- `backend_tests/backend_test.py` -> Hits /handled, /unhandled/, + /checkout backend demo APIs
+
+create_cron_job.sh -> GCP-cron job (runs every 20 min from midnight-6am) -> TravisCI (runs tests)
+
+# Tests
+
+## Setup
+python -m venv env
+source env/bin/activate
+pip install -r requirements.txt
+
+## FrontEnd / Selenium (`frontend_tests` directory)
+Pulls up Sentry frontend in various browsers in parallel via selenium scripts.
+Test case will add items to cart and then click checkout
+
+```
+py.test -s -n 2 frontend_tests
+```
+
+`-n` is for number of threads
+
+
+# Additional Setup instructions/context
+### Python2
+
+SAUCE_USERNAME + SAUCE_ACCESS_KEY environment variables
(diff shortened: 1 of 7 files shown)


thinkocapo:
@ndmanvar can you link to the transactions in Sentry.io so I can check?

thinkocapo:
@ndmanvar should we maybe give the test a new name? Otherwise it's hard to tell what's running on Empower Plant vs Tool Store. All those metric alerts I got (emailed you) appeared like they were from Empower Plant.

I suppose we could keep them in place as we decommission ToolStore, but it could definitely create confusion in the near term. Maybe i'll rename them in a next PR.

thinkocapo:
@ndmanvar one more thought, you're sure you want test data automation in application-monitoring repo and not the test-data-automation repo? You have everything in a /tests sub-directory but wanted to ask anyways.

thinkocapo (commented):
see all the comments I left leading up to this review

tests/README.md
```diff
@@ -0,0 +1,72 @@
...
+- `.travis.yml` -> runs automated tests / simulations
+- `conftest.py` -> Sauce Labs configuration (browsers) for frontend_tests
+- `backend_tests/backend_test.py` -> Hits /handled, /unhandled/, + /checkout backend demo APIs
+
+create_cron_job.sh -> GCP-cron job (runs every 20 min from midnight-6am) -> TravisCI (runs tests)
+
+# Tests
+
+## Setup
+python -m venv env
```
thinkocapo, line 15: good, this is critical or else it could pollute the global space on the VM, or even someone's virtual env that they're using for application monitoring.

We will keep using virtual environments as this is askin to using an individual node_modules directory and cuts down on the potential for SE's having errors due to this stuff. Localizes it.

thinkocapo:
@ndmanvar one thing I want to get down perfectly from start, and maybe we can do it on a call together, or I'd have to run your code/tests...

## Question
What are all transactions that get created? 

## Why
We need to get a good grip on what to expect (expected behaviors. After all, this is Testing we're doing) because it's super confusing later on when we suspect things aren't working. I don't want to be eyeballing graphs in Discover with a subjective sense of "looks like they're appearing..." I'd rather say, "It's been running 24hrs so I'm expecting to see +10,000 total transactions..." like a real test. Because small changes to our application-monitoring repo can easily impact this and we need to be aware ASAP and not notice it +6 months later, because then it's a headache to fix (so much has changed by that point).

## For Example
_Successful checkouts_ re-route to the Order Complete page, and this creates a navigation transaction. Is this getting captured? Creative team made this page because we wanted more transactions. So if this is not getting captured, then that's a problem.

_Unsuccessful checkouts_ re-route to the Error page, so does this transaction get captured as well?

## Conclusion
By having a base case down, then we can really predict how many thousands or millions of transactions + errors we expect to see as we scale this thing up (i.e. increase velocity of running TDA). I suppose right now we're not really controlling this (as it just repeats on a while loop) but long-term there needs to be more, and still need to know this for the reasons I stated above.

PR #18, opened 2021-07-12
query params trigger alert

thinkocapo:
## Overview
- Set `se=<name>` tag on /products pageload so that it's retained for when the error gets made later.
- Verify it stays on scope across navigation state changes.

NOTE - email gets set in src/index.js and overrides the email in the checkout form (which gets passed in the request body). Email normally provided by session login or signup/checkout, and not right when a visitor lands on the home page.

NOTE - Getting the 'se' tag off of the sentry scope and having to set for every request's headers... could implement a new requests client so can hook into the request to set this (DRY).

NOTE - checking for 'se' tag in python as None or "undefined" because it's sometimes sent as "undefined" from frontend. We implement the change here in the before_request, or else have to instrument logic for every frontend request's 'header' parameter.

## Steps
1. visit the products page with your name as a query param for `se` http://localhost:5000?se=will
2. This gets parsed in componentDidMount "?se=will" and set as a tag called 'se'.
3. the tag's value persists through healthy and unsuccessful checkouts.
4. it triggers the Alert.

Must have Alert set like:
image

## Testing
The transactions are...

[/products](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:007d3d98dd374a17801b9d00377f24f0/?field=title&field=event.type&field=project&field=se&field=timestamp&name=All+Events&project=5808623&query=&sort=-timestamp&statsPeriod=1h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) has se tag.

all the way through...

[/error](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:f99f704208b243518e82ca7f3398d4cc/?field=title&field=event.type&field=project&field=se&field=timestamp&name=All+Events&project=5808623&query=&sort=-timestamp&statsPeriod=1h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) transaction and the [error](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:164a99dd65604f67b193e87b48b8e795/?field=title&field=event.type&field=project&field=se&field=timestamp&name=All+Events&project=5808623&query=&sort=-timestamp&statsPeriod=1h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) both have se tag.

And the SE tag in the alert looked like:
image
^ you could remove this tag from your alert configuration.

The Alert did not come through when I left out the `?se=will` query param, which is the expected behavior.

Also...
If you don't put a `se=<name>` query param, then the `se` tag will not get set:
image

## Didn't Work
Clicking 'Browse Products' doesn't set it for you. You have to remember to set it manually when you're on the /products page. You could optionally pass one via a .env file property, and set it if it's present, but that may be too much work AND that's only useful if you're running it locally. On the admin instance - there'd be know way what to auto set that as, and so it's better to write it manually in the URL bar.

https://softchris.github.io/books/react/router-parameters/ did not work because "react React Hook "useQuery" cannot be called in a class component. React Hooks must be called in a React function component or a custom React Hook function"

Because this was per route, and we solutioned something different in src/index.js to apply everywhere.
```
    let query = this.props.history.location?.search 
    query ? 
      Sentry.configureScope(scope => {
        scope.setTag("se", query.split("se=").pop())
      })
      :console.log("no query string")
```

### DIFF
--- a/src/components/Checkout.js
+++ b/src/components/Checkout.js
@@ -47,6 +47,10 @@ class Checkout extends Component {
 
     let response = await fetch(`${BACKEND}/checkout`, {
       method: "POST",
+      headers: {
+        "se": "will"
+        // "email": this.state.email // email here is already in the body
+      },
       body: JSON.stringify({
         cart: cart,
         form: this.state
--- a/src/components/Checkout.js
+++ b/src/components/Checkout.js
@@ -45,17 +45,15 @@ class Checkout extends Component {
     // Do this or the trace won't include the backend transaction
     Sentry.configureScope(scope => scope.setSpan(transaction));
 
-    let se
+    let se, customerType, email
     Sentry.withScope(function(scope) {
-      se = scope._tags.se
-      console.log("scope._tags.se", se)
+      [ se, customerType ] = [scope._tags.se, scope._tags.customerType ]
+      email = scope._user.email
     });
 
     let response = await fetch(`${BACKEND}/checkout`, {
       method: "POST",
-      headers: {
-        'se': se
-      },
+      headers: { se, customerType, email },
       body: JSON.stringify({
         cart: cart,
         form: this.state
@@ -65,8 +63,6 @@ class Checkout extends Component {
       console.log("> catches error", err)
       throw Error(err) 
     })
-
-    console.log("> response", response)
     console.log("> ok | status | statusText", response.ok, response.status, response.statusText)
 
     if (!response.ok) {
--- a/src/components/Checkout.js
+++ b/src/components/Checkout.js
@@ -45,7 +45,7 @@ class Checkout extends Component {
     // Do this or the trace won't include the backend transaction
     Sentry.configureScope(scope => scope.setSpan(transaction));
 
-    let se // TODO try configureScope so it persists after the error?
+    let se
     Sentry.withScope(function(scope) {
       se = scope._tags.se
       console.log("scope._tags.se", se)
--- a/src/components/Checkout.js
+++ b/src/components/Checkout.js
@@ -45,10 +45,16 @@ class Checkout extends Component {
     // Do this or the trace won't include the backend transaction
     Sentry.configureScope(scope => scope.setSpan(transaction));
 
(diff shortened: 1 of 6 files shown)


thinkocapo:
tested this on the admin instance (App Engine) and it worked fine

Re-tested /product/:id /employee/:id and /product-join workflow, all worked fine.

src/index.js
```diff
@@ -159,20 +159,10 @@ class App extends Component {
...
-                ></SentryRoute>
-                <Route path="/products">
-                  <Products />
-                </Route>
-                <Route path="/products-join">
-                  <ProductsJoin />
-                </Route>
+                <SentryRoute path="/employee/:slug" component={Employee}></SentryRoute>
+                <SentryRoute path="/product/:id" component={Product}></SentryRoute>
+                <Route path="/products" component={Products} />
```
thinkocapo, line 164: passing the Products component into the Route's tag like this is 100% required, or else `this.props.history` is always null.

Can't do:
```
 <Route path="/products">
                  <Products />
                </Route>
```
Must do:
```
<Route path="/products" component={Products} />
```

ndmanvar (approved):

thinkocapo:
se, customerType, email now persist across frontend and backend. They're all set from src/index.js and passed via headers
@ndmanvar @idosun 
image

src/components/Checkout.js
```diff
@@ -45,8 +45,15 @@ class Checkout extends Component {
...
 
+    let se, customerType, email
+    Sentry.withScope(function(scope) {
+      [ se, customerType ] = [scope._tags.se, scope._tags.customerType ]
+      email = scope._user.email
+    });
+
     let response = await fetch(`${BACKEND}/checkout`, {
       method: "POST",
+      headers: { se, customerType, email },
```
thinkocapo, line 56: I recorded a TODO for using a requests library where we can hook into it to set these headers, so we put all the code there for parsing the tags, email, and setting it as headers too.

thinkocapo:
## Testing
works in App Engine
image

PR #20, opened 2021-07-19
TDA Products Join

thinkocapo:
Test that you can run TDA on the /products-join route and make any necessary updates.

#### Overview
This worked, the test is only getting the /products-join endpoint, it does not add items to the shopping cart, but that's okay.
```
driver.get("http://application-monitoring-react-dot-sales-engineering-sf.appspot.com/products-join")
```

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,6 +1,6 @@
   
 env: standard
-service: application-monitoring-react
+service: -<SERVICE_NAME_HERE>-
 runtime: nodejs14
 
 handlers:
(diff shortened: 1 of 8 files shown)


ndmanvar (approved):

ndmanvar:
looks good + straightforward!

thinkocapo:
@ndmanvar Tested - I ran it from the compute engine, it produced transactions for /products and /products-join in near equal amounts, which was expected.
image

app.yaml
```diff
@@ -1,6 +1,6 @@
   
 env: standard
-service: application-monitoring-react
+service: -<SERVICE_NAME_HERE>-
```
thinkocapo, line 3: this change helps prevent people from accidentally overwriting the master admin instance.

thinkocapo:
Noting the difference in `products` vs `products_join` after running TDA for nearly 24 hours
image

PR #24, opened 2021-07-30
Refactor - Release Health, Pytests, Project Structure

thinkocapo:
## Todo
- rename tests so doesn't conflict with ToolStore's, when viewing in Saucelabs
- prod app.yaml should be hitting https not http.
- `/list` app needs more errors, and TDA pytest for hitting it.

## Done
- `crash=true` and `crash=1.0` now accepted, pytest sends a random value for crash rate.
- updated pytest so cart checkout is sometimes successful, so `/complete` transactions appear
- unique fingerprint'ing for Release Health
- `try/except` block for Pytests, around button clicking. gracefully handle (i.e. don't fail the whole pytest/selenium in saucelabs).
- updated javascript and python sdk's to `6.10.0` and `1.3.1`.
- /react directory for react app, relocated from repo root
- loader icon for /products /products-join
- added delays to waterfall spans for /products
- `crash=true` is processed by src/index.js so you can pageload from any place in the EmpowerPlant (as long as it's a page load/refresh) with crash=true, and get the unhandled error that way.
- `crasher()` function takes care of probability() and unhandled errors, can re-use it from anywhere you want. Made a separate pytest for keeping crashes to the homepage, so that the selenium tests that do shopping cart purchases will 100% always complete (and not crash early). Release Health crash rates will still fluctuate accordingly. Separate test means good separation of concerns. We get higher total volume this way, as the pytest for /checkout won't get disrupted (causes +3 missed events).
- `beforeSend` sets custom fingerprints for errors from `crasher()`. see this [transaction's](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:9009f2ed45d343a79dc895b8393f90ac/?field=issue&field=title&field=event.type&field=project&field=timestamp&name=All+Events&project=5808623&query=url%3Ahttp%3A%2F%2Flocalhost%2A&sort=-timestamp&statsPeriod=1h&user=408003) grouping info, it shows the fingerprint that's set in `beforeSend`.
- Adjusted `test_add_to_cart` so that it gives pageload transactions which are far more complete (includes xhr's, resource.images, react lifeyccle methods). It will still complete the shopping cart purchase. If you open [these transactions](https://sentry.io/organizations/testorg-az/discover/results/?field=title&field=event.type&field=project&field=transaction.op&field=timestamp&name=All+Events&project=5808623&query=transaction%3A%22%2Fproducts%22+url%3A%22http%3A%2F%2Fappl%2A%22&sort=-timestamp&statsPeriod=24h), notice they weren't doing a full pageload's worth of spans (that's bad)
- Updated pytest so that `/products` transactions are `pageload` and not `navigation` (has less data)

## Testing
`staging` instance used for testing.
`production` instance right now is reserved for the 'Refactor - LCP Investigation' PR

1. run 24hr of pytest with crash=true on homepage, see if Crash Rates in Release Health start lowering. 1 New Issue per release.
2. deploy to prod, run TDA from VM once, then continually and auto-deploy the VM instance

## Demo Notes
"Homepage, with no xhr requests, what you get out of the box..."


thinkocapo:
Transition from 7.4 to 8.1 in 24hr view doesn't look good (7.4 needs data from +24hrs ago)
Transition from 7.4 to 8.1 in 7hr view
image
Transition from 7.4 to 8.1 in 7day view
image

thinkocapo:
## Testing
No sleep timeout

Pytest
1Failed, 11 Passed, 4 Warnings

295 total tx's
image

react/src/index.js
```diff
@@ -86,22 +91,29 @@ class App extends Component {
       }
     };
 
+    // new PerformanceObserver(entryList => {
```
thinkocapo, line 94: this was for debugging Largest Contentful Paint not getting set, but leaving it in somehow caused errors once it was in prod + tda was running

leave as commented as we're still fixing LCP, may need it.

release.sh
```diff
@@ -1,6 +1,7 @@
 # Sets a release according to Calendar Versioning. Month, followed by week of the month.
 # In sentry-demos/tracing, we did not manage this whatsoever
-# This release.sh is re-used in many places so that we don't need 3 different Makefiles like in sentry-demos/tracing
+# /react and /flask both have a run.sh script that utilizes this code
+# This release.sh is used by deploy.sh, react/run.sh, flask/run.sh
```
thinkocapo, line 4: so we don't need to maintain 3 different Makefiles.

PR #25, opened 2021-08-06
Error Handling Fetches

thinkocapo:
## Todo
Update the ProductsJoin
- `/list` route
- `/about`
- `/employee/:id`
- try statement loop until /products are loaded

# Done
- `try/catch` for /products
- fixed fingerprint homepage error (Release Health), since it was having a different stacktrace per browser.
- updated error handling for /checkout should bring more consistent errors via TestDataAutomation.

Firefox and Chrome both error'd into same code block:
image

### DIFF
--- a/flask/app.yaml
+++ b/flask/app.yaml
@@ -1,2 +1,2 @@
-service: -<NAME>-
+service: staging-application-monitoring-flask
 runtime: python38
--- a/flask/db.py
+++ b/flask/db.py
@@ -35,7 +35,8 @@ else:
 
 # N+1 because a sql query for every product n
 def get_products():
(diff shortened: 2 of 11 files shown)


thinkocapo:
Looks unhandled according to code, but says handled:true

because...
```
    if (!response.ok) {
      throw new Error(response.status + " - " + (response.statusText || "Internal Server Error"))
    } else {
      const jsonValue = await response.json()
      return Promise.resolve(jsonValue)
    }
```
end up here:
```
    try {
      var result = await this.checkout();
    } catch(err) {
      console.log("*CAUGHT ERR*", err)
      Sentry.captureException(err);
    }
```
image

PR #28, opened 2021-08-17
Add test for /products pageload

ndmanvar:
https://app.saucelabs.com/tests/a1d40f606b4c4bac987c2e8ff02f58c6

### DIFF
--- a/tests/frontend_tests/test_products.py
+++ b/tests/frontend_tests/test_products.py
@@ -1,0 +1,36 @@
+import pytest
+import time
+import yaml
+import random
+import sentry_sdk
+
+# This test is for the homepage '/' transaction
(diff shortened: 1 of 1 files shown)


thinkocapo:
@ndmanvar the `test_add_to_cart` was refactored in https://github.com/sentry-demos/application-monitoring/pull/24 so that it causes a pageload transaction for /products. This is how the pytest for /products-join works too.

PR mentions "Updated pytest so that /products transactions are pageload and not navigation (has less data)"

Before that, /products-join was a pageload and /products was a navigation. But now, they're both pageload.

Can you review again the 4 tests in /tests and see if those are sufficient?

ndmanvar:
yep, you are correct- those tests are sufficient. closing out

PR #30, opened 2021-08-30
Throw random error on forced crash

cstavitsky:
**Background**
Functionality already exists to force a crash to the demo application by passing the query param `crash=true`. Previously this would result in only one type of error, `UnhandledException`. Given there was only one error type, only one Issue in Sentry was created and it was therefore less realistic for demo purposes than being able to show a customer multiple errors.

**Updates in this PR**
- This PR adds a handful of other error types and chooses a random one to throw, when a crash is forced.
- The option to crash based on probability (i.e. 20% of the time with `crash=0.2`) should remain unchanged.
- See screenshot of sentry.io showing different Issue entries for each error type. Note that a `beforeSend` hook currently forces all `UnhandledException`s to be set to the same fingerprint of `'unhandled-exception'`. This PR doesn't touch that logic because I'm not sure if/how we'll want the fingerprinting logic to change. Without any changes to the current fingerprinting logic, each of these errors should be grouped according to stack trace, as seen in the screenshot below.

## Link to Issues
- 💡 now showing `handled: false`, and with source maps uploaded.
- [Unhandled error that existed previously](https://sentry.io/organizations/testorg-az/issues/2614163848/?project=5938093&query=is%3Aunresolved)
- [RangeError](https://sentry.io/organizations/testorg-az/issues/2616426553/?project=5938093&query=is%3Aunresolved)
- [SyntaxError](https://sentry.io/organizations/testorg-az/issues/2616426437/?project=5938093&query=is%3Aunresolved)
- [URIError](https://sentry.io/organizations/testorg-az/issues/2616435052/?project=5938093&query=is%3Aunresolved)
- [TypeError](https://sentry.io/organizations/testorg-az/issues/2616435066/?project=5938093&query=is%3Aunresolved)

## Screenshots

Screen Shot 2021-08-30 at 4 30 59 PM

Screen Shot 2021-08-30 at 4 30 17 PM

Screen Shot 2021-08-30 at 4 30 25 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,15 @@ Also called the Empower Plant UI/UX. This project was bootstrapped with [Create 
 Permit your IP address in CloudSQL.
 
 **Test**
-1. Create a .env and enter your DSN. See .env.example for an example.
-2. Create a flask/.env and enter your DSN. See .env.example for an example. Fill out all fields so data can be read from the database.
+1. Create a `react/.env` and enter your DSN. See .env.example for an example.
+2. Create a `flask/.env` and enter your DSN. See .env.example for an example. Fill out all fields so data can be read from the database.
 
 **Production** - AppEngine
 1. Enter a value for REACT_APP_BACKEND in .env, as this represents the Flask AppEngine.
 
 ```
 # React
+cd react
 npm install
 
 # Flask
@@ -41,14 +42,23 @@ cd flask
 
 **Test**
 ```
-# React w/ sourcemaps, suspect commits
+# Pick one of two ways to run the React app:
+
+# 1) Run React app w/ sourcemaps, suspect commits
+# Recommended to use this command rather than `npm start`, as
+# run.sh uploads source maps and handles crashes more
+# realistically.
 ./run.sh
 
-# React w/ hot reload
+# 2) Run React app w/ hot reload
+# NOTE: this will cause crashing errors to be tagged
+#   in sentry as handled (`handled: true`)
 npm start
 
+```
 
-# Flask
+```
+# Run the Flask app
 ./run.sh
 ```
 
--- a/react/src/utils/errors.js
+++ b/react/src/utils/errors.js
@@ -1,22 +1,49 @@
 import { createBrowserHistory } from 'history';
 const history = createBrowserHistory();
 
+// ERRORS
+const notAFunctionError = () => {
+  const someArray = [{ func: function () {}}];
+  someArray[1].func();
+}
+const uriError = () => decodeURIComponent('%');
+const syntaxError = () => eval('foo bar');
+const rangeError = () => {
+  throw new RangeError('Parameter must be between 1 and 100');
+}
+const unhandledError = () => {
+  throw new UnhandledException('unhandled error')
+}
+
+const randomErrors = [
+  notAFunctionError,
+  uriError,
+  syntaxError,
+  rangeError,
+  unhandledError
+]
+
+const throwRandomError = () => {
+  const randomNum = parseInt(Math.random()*randomErrors.length)
+  randomErrors[randomNum]()
+}
+
 var probability = function(n) {
   return !!n && Math.random() <= n;
 };
 
 const crasher = () => {
-  let queryParams = new URLSearchParams(history.location.search)
+  const queryParams = new URLSearchParams(history.location.search)
   if (queryParams !== "") {
-      let crash = queryParams.get("crash")
-      if (crash) {
-            console.log("> crash", crash)
-            if (crash === "true" || probability(parseFloat(crash))) {
-              throw new UnhandledException('unhandled error')
-            }
+    const crash = queryParams.get("crash")
+    if (crash) {
+      console.log("> crash", crash)
+      if (crash === "true" || probability(parseFloat(crash))) {
+        throwRandomError();
       }
+    }
   } else {
-      console.log("> queryParam was", queryParams)
+    console.log("> queryParam was", queryParams)
   }
 }
 
--- a/react/src/utils/errors.js
+++ b/react/src/utils/errors.js
@@ -6,7 +6,9 @@ const notAFunctionError = () => {
   const someArray = [{ func: function () {}}];
   someArray[1].func();
 }
-const uriError = () => decodeURIComponent('%');
+const referenceError = () => {
+  throw new ReferenceError('undefinedVariable is not defined')
+};
 const syntaxError = () => eval('foo bar');
 const rangeError = () => {
   throw new RangeError('Parameter must be between 1 and 100');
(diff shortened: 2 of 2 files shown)


thinkocapo:
@cstavitsky looks good. Can you put a link to your Issues page or the individual Issues so I can check these? Screenshots are good but most important is the end piece that we see in Sentry.io

thinkocapo:
@cstavitsky I found your Project but please share the link here to the Issues dashboard or individual issues anyways. Can you upload source maps? I don't see any deminified stacktraces. Looking to make sure it shows the error in the stacktrace. We want to see TypeError and not throwRandomError, or else it'd put to waste having the UnhandledRejections as I know @ndmanvar likes to show that. He wants to see Unhandledrejection in the stacktrace.

Good job not touching the beforeSend logic for grouping. That grouping is to ensure there's at least +1 New Issue in every release, or else in Release Health it would show 0 New Issues. Note the fingerprint is based on unhandled-rejection plus the release number (which rotates every week, hence your'e guaranteed New Issue every week). SE:tag does not guarantee a new issue (because when you se:tag and /checkout an error, it's the same exact error + fingerprint you did the previous week/month already). 

I'm not sure that logic still gets reached, it probably does - but I'd need to see the link to your UnhandledRejection in order to decipher that.

thinkocapo:
@cstavitsky here's that fingerprint UnhandledException. The fingerprint is not unhanlded-exception, it's unhandled-exception plus the Release number.
```
    if (exception instanceof UnhandledException) {
      event.fingerprint = ['unhandled-exception', process.env.REACT_APP_RELEASE ];
```

If you really wanted to, you could make your exceptions inherit from a BaseClass and we could change this to something like
```
    if (exception instanceof BaseException) {
      event.fingerprint = [exception.type, process.env.REACT_APP_RELEASE ];
```
This would ensure you get +5 New Issues per Release (1 for every exception type you added). @ndmanvar 

This probably isn't clear and it's a lot of testing. I'll probably take it on myself.

thinkocapo:
also are all of these handled:false errors? I thought Neil wanted to be able to show unhandled errors. Are we breaking that? Are you still able to 100% show an unhandled error? Once you link to these errors, we can check each one. `crash` denotes unhandled and I'm pretty sure all of yours are unhandled.

thinkocapo:
@cstavitsky one final thought, are each of these errors producible in Chrome, Safari and Mozilla? I could check the Mozilla documentation. They probably are (as they sound common), but once we start automating Pytests via Selenium on these (`crasher` is used on the homepage, for powering Release Health) then we'll find out fast. Worst case, it just produces some unrecognizable error. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError#browser_compatibility

Had problems adding error types in the past because they don't all occur/throw the same across browsers. Trying to avoid things sneaking in and surprising us in Prod.

No action needed here, but something to be aware of in the future.

cstavitsky:
@thinkocapo jotting down a few responses:

> @cstavitsky I found your Project but please share the link here to the Issues dashboard or individual issues anyways. Can you upload source maps? I don't see any deminified stacktraces. Looking to make sure it shows the error in the stacktrace. We want to see TypeError and not throwRandomError

- done -- uploaded source maps and updated the PR description with link to individual issues. From my current understanding it shows the desired error in the stacktrace, but if there's something that still needs work just let me know.

> also are all of these handled:false errors? I thought Neil wanted to be able to show unhandled errors. Are we breaking that? Are you still able to 100% show an unhandled error? Once you link to these errors, we can check each one. crash denotes unhandled and I'm pretty sure all of yours are unhandled.

- From the Issue links I added to the PR you can see that they are now showing `handled: false`. I asked Neil about this and he pointed out that mine were showing `handled: true` because I was using `npm start` (apparently the framework handles these errors rather than actually crashing) rather than your `./run.sh` script. So I believe those work as expected.

> The fingerprint is not unhanlded-exception, it's unhandled-exception plus the Release number.

Ahhh... Missed that the release number was tacked on, thanks for pointing that out. Would it be bad to have all errors (and not just the forced-crash ones) show up as 'new' each release? Because if not, we could just drop the conditional check for `UnhandledException` and do something like this...?

```
beforeSend(event, hint) {
  const exception = hint.originalException
  event.fingerprint = [exception.stack, process.env.REACT_APP_RELEASE ];
  return event;
}
```

thinkocapo (changes requested):
Nice updates, all looking good. Can you double check your stacktraces for uriError again?

Try clicking Original vs Minified. It appears unminified when you click Original but I'm not sure why the code isn't showing. I can't recall if I've seen this happen before.

If it turns out there's a strange reason it won't work then maybe we can remove it or replace it with a different error type.
image

cstavitsky:
> Nice updates, all looking good. Can you double check your stacktraces for uriError again?
> 
> Try clicking Original vs Minified. It appears unminified when you click Original but I'm not sure why the code isn't showing. I can't recall if I've seen this happen before.
> 
> If it turns out there's a strange reason it won't work then maybe we can remove it or replace it with a different error type.
> image

@thinkocapo good question.

## tl;dr / outcome

I suggest we replace `URIError` with a different error type. It technically works, but the reason why is likely to be confusing for SEs and customers looking at the demo who don't have time to dig into what's going on. 

I pushed a commit replacing it with `ReferenceError` - https://sentry.io/organizations/testorg-az/issues/2618639686/?project=5938093&query=is%3Aunresolved

## Explanation of What I Think's Happening with URIError

I looked into it a bit and it looks like for the [URIError issue page](https://sentry.io/organizations/testorg-az/issues/2616892928/?project=5938093&query=is%3Aunresolved), even though the stack trace doesn't appear automatically, the stack trace is expandable for source code after the first line of the stack trace (see screenshots at bottom, showing source code that appears after clicking the button to expand).

When I examine the stack trace for this error vs the others that show up fine, my current assumption is that this occurs because the [actual error occurs in the JavaScript built-in function `DecodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent). Since that final line of the stack trace is in the JS language and not actually in our codebase, it would have nothing to display from our source code. In contrast, the final line of the stack trace for all the other errors is actually in our codebase.

However, I did notice that on the `URIError` issue page in sentry.io it does show 2 problems processing the event, whereas the other errors show one. I just wanted to point this out in case you know why that might be happening.

Screen Shot 2021-09-01 at 9 12 55 AM

## Reference Screenshots

Screen Shot 2021-09-01 at 9 06 32 AM

Screen Shot 2021-09-01 at 9 05 42 AM

thinkocapo (approved):

thinkocapo:
@cstavitsky I approved, go ahead and merge.

Nice find on
> the stack trace is expandable for source code after the first line of the stack trace 
I see what you mean. Same for how decodeURI is not really our codebase, so it doesn't show.

It's still good you switched to ReferenceError because the decodeURI thing is easily forgettable and would mean we have to educate people on it, it also may throw audiences off.

Nice work!

PR #32, opened 2021-09-13
Route requests to different backend URLs based on backend-type query parameter

cstavitsky:
### Task Overview
- Permit hitting different backend URLs based on a passed query parameter, i.e. `?backend=express`. 
- This will be used soon by me and Simon for our Springboot/Express onboarding tasks
- This is very easily extensible to any other type of backend we might want empower plant to support

### Acceptance Criteria
✅ Should show up in `tags` in sentry.io
✅ Should hit different URLs depending on environment and specified backend type
✅ Should default to flask if no backend specified
✅ If an invalid backend type is specified, i.e. `?backend=invalid_type`, `alert` the user and proceed with using the flask backend.
✅ This should work whether accessing the homepage first and clicking through to products page, or, accessing the products page directly.

### Screenshots
You can see that the `backendType` was reported to Sentry as a tag.

Screen Shot 2021-09-13 at 11 04 42 AM

### DIFF
--- a/react/src/index.js
+++ b/react/src/index.js
@@ -7,6 +7,7 @@ import { Integrations } from '@sentry/tracing';
 import { createBrowserHistory } from 'history';
 import { Router, Switch, Route } from 'react-router-dom';
 import { crasher, UnhandledException } from './utils/errors'
+import { determineBackendType, determineBackendUrl } from './utils/backendrouter'
 
 import { Provider } from 'react-redux'
 import { createStore, applyMiddleware, compose } from 'redux'
@@ -43,6 +44,8 @@ if (window.location.hostname === "localhost") {
 } else { // App Engine
   ENVIRONMENT = "production"
 }
+
+let BACKEND_URL
 const DSN = process.env.REACT_APP_DSN
 const RELEASE = process.env.REACT_APP_RELEASE
 console.log("ENVIRONMENT", ENVIRONMENT)
@@ -76,7 +79,7 @@ const store = createStore(
 )
 
 class App extends Component {
-  
+
   constructor() {
     super();
     this.state = {
@@ -96,9 +99,13 @@ class App extends Component {
 
     let queryParams = new URLSearchParams(history.location.search)
 
+    // Set desired backend
+    let backendTypeParam = new URLSearchParams(history.location.search).get("backend")
+    const backendType = determineBackendType(backendTypeParam)
+    BACKEND_URL = determineBackendUrl(backendType, ENVIRONMENT)
+
     // These also get passed via request headers
     Sentry.configureScope(scope => {
-      
       const customerType = ["medium-plan", "large-plan", "small-plan", "enterprise"][Math.floor(Math.random() * 4)]
       scope.setTag("customerType", customerType )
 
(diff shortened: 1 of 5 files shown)


react/src/index.js
```diff
@@ -130,14 +139,20 @@ class App extends Component {
                 <Route exact path="/" component={Home} />
                 <Route path="/about" component={About} />
                 <Route path="/cart" component={Cart} />
-                <Route path="/checkout" component={Checkout} />
+                <Route path="/checkout">
+                  <Checkout backend={BACKEND_URL} history={history} />
```
cstavitsky, line 143: Since we are no longer passing `Checkout` as a `component`, we lose out on the inclusion of `history` by default. 

- docs link: https://reactrouter.com/web/api/Route/route-props

Of the three routes I changed, this only affects the `/checkout` route.

react/src/index.js
```diff
@@ -130,14 +139,20 @@ class App extends Component {
...
+                </Route>
                 <Route path="/complete" component={Complete} />
                 <Route path="/error" component={CompleteError} />
                 <Route path="/cra" component={Cra} />
                 <SentryRoute path="/employee/:slug" component={Employee}></SentryRoute>
                 <SentryRoute path="/product/:id" component={Product}></SentryRoute>
-                <Route path="/products" component={Products} />
-                <Route path="/products-join" component={ProductsJoin} />
+                <Route path="/products">
```
cstavitsky, line 150: In general the proposed pattern here is, instead of calculating the backend URL in each of 3 individual files, [for example, here](https://github.com/sentry-demos/application-monitoring/blob/31ae8f6f7b946205e7d3dd88326a9fd799dd8b51/react/src/components/Products.js#L10-L16), we just calculate it once and pass it as `props` in our `Route` definitions.

react/src/utils/backendrouter.js
```diff
@@ -0,0 +1,38 @@
+const DEFAULT_BACKEND = "flask"
+const SUPPORTED_BACKEND_TYPES = {
```
cstavitsky, line 2: This config for supported backend types could potentially be put into a config file if we plan to add more backend types. But it seemed like overkill to do that now.

react/src/index.js
thinkocapo: so you couldn't do
`<Route path="/checkout" backend={BACKEND_URL} component={Checkout} />`? Wouldn't that be less refactoring? But maybe Checkout wouldn't get access to this.props.backend that way. Not sure.

I remember having to decide or evaluate once the two different ways of doing this...I can't remember if one way caused a loss of functionality or not, but as long as the transactions are still getting full context and parameterizing correctly then we should be good. The ones with `<SentryRoute>` are different.

cstavitsky: Yeah let me double check if we can do it that way and report back - i tried a few different things and don't remember if that was one of them. If we can do it that way they I support it.

cstavitsky: @thinkocapo I double checked this: `<Route path="/checkout" component={Checkout} backend={BACKEND_URL} />`. 

From within the `Checkout` component, `this.props.backend` appears not to have been passed correctly, as it shows up as `undefined`. So as far as a path forward, we could go with the way I've done it here, or, if you have other suggestions I'm open to them.

thinkocapo (approved):

PR #35, opened 2021-10-11
import new framework that supports mobile-native. frontend_tests -> d…

ndmanvar:
`frontend_tests` -> `desktop_web`


ndmanvar:
looks like there is a bug where we can't scroll down in on the app in android when on the checkout/place-order screen (I'll show you tomorrow @thinkocapo )
image

PR #46, opened 2021-11-17
Update tests to point to React Native release 1.9

cstavitsky:
- Point to the most recent release: https://github.com/sentry-demos/sentry_react_native/releases/tag/1.9

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -59,7 +59,6 @@ def pytest_addoption(parser):
 def data_center(request):
     return request.config.getoption('--dc')
 
-
 @pytest.fixture(params=desktop_browsers)
 def desktop_web_driver(request, data_center):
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -125,7 +124,7 @@ def android_emu_driver(request, data_center):
         'deviceName': 'Android GoogleAPI Emulator',
         'platformVersion': '10.0',
         'platformName': 'Android',
-        'app': "https://github.com/sentry-demos/sentry_react_native/releases/download/1.8/app-release.apk",
+        'app': "https://github.com/sentry-demos/sentry_react_native/releases/download/1.9/app-release.apk",
         'sauce:options': {
```
cstavitsky, line 128: I'd like to update this so that we fetch a list of releases for the `sentry_react_native` repo via the Github API, and select the most recent one -- so that we don't have to tick this value manually anymore. 

However, don't want to slow this down so I can work on that separately.

thinkocapo: ok sounds good

thinkocapo (approved):

PR #47, opened 2021-11-17
Two scripts for mobile and web in parallel

thinkocapo:
It worked:
https://www.loom.com/share/9df1f579133049868af4f68765277769

```
(envdesktop) <>@<>:~/application-monitoring/tests$ cat script1.sh 
while true; do clear && pytest -s -n 4 mobile_native; done

(envdesktop) <>@<>:~/application-monitoring/tests$ cat script2.sh 
while true; do clear && pytest -s -n 4 desktop_web; done
```
The data no longer has gaps, the data is fuller. This is because the mobile_native tests are being run in parallel to the desktop_web tests, on different cores.
image


tests/README.md
```diff
@@ -41,6 +41,7 @@ Use an isolated VM since it's constantly occupying +2 threads simultaneously
 ```
 source .virtualenv/bin/activate
 nohup ./script.sh &
+nohup ./script.sh >/dev/null 2>&1 &
```
cstavitsky, line 44: I think I'm misunderstanding something with the instructions here. Can you explain why the readme is stating to run the same script (`script.sh`) twice?

- I understand the first call (L43) is running `script.sh` with no-hangup, and doing so in the background.
- And I understand that the second call (L44) is doing the same thing except discarding anything output to stdout.

But I'm not following why we run the same script, `script.sh`, twice? 

Or is the intent that you'd run `script1.sh` and `script2.sh` at the same time?

tests/script2.sh
```diff
@@ -0,0 +1 @@
+while true; do clear && pytest -s -n deskto_web; done
```
cstavitsky, line 1: `deskto_web` <--- is this a typo for `desktop_web`?

cstavitsky (commented):
Couple questions

tests/script2.sh
cstavitsky: Also, can we rename `script1.sh` to `mobile_native.sh` and rename `script2.sh` to `desktop_web.sh`?

tests/README.md
thinkocapo: The instructions are misleading. I'm updating it now

tests/script2.sh
thinkocapo: I was thinking that too, so I've updated the names of the scripts.

Yes typo. I had tested this from the compute engine but didn't check code into source control there, so that's why it worked there. Committing the code via my local machine. Updated.

tests/README.md
```diff
@@ -40,7 +40,8 @@ py.test -s -n 4 desktop_web
 Use an isolated VM since it's constantly occupying +2 threads simultaneously
 ```
 source .virtualenv/bin/activate
-nohup ./script.sh &
+nohup ./script1.sh >/dev/null 2>&1 &
```
cstavitsky, line 43: should this be 

```
nohup ./mobile_native.sh >/dev/null 2>&1 &
nohup ./desktop_web.sh >/dev/null 2>&1 &
```

?

cstavitsky: (change name from script1/script2)? ^

cstavitsky (approved):
LGTM 👍

PR #48, opened 2021-12-01
Environment Variables for Express and SpringBoot Back Ends, Updates

thinkocapo:
## Done
- fixed Environment Variables for Express and SpringBoot Back Ends
- placeholder values in express/example.env so it's clear there's something needed there.

This was tested via the Cloud Run served React + Express + Springboot apps. Previously, this was tested via a React App served on a laptop's localhost.

All of these were tested and now work:
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/products?backend=express
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/products?backend=springboot
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com?backend=express
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com?backend=springboot

[Adding Custom Environment Variables](https://create-react-app.dev/docs/adding-custom-environment-variables/)
image

## Also Done
- sampled out Express.js transactions for pre-flight requests, via `tracesSampler`
- added README section on Deploying to Staging


README.md
```diff
@@ -59,12 +59,19 @@ Add +2 quantity of a single item to Cart and purchase in order to trigger an Err
 
 
 ## Deploy
-This run script deploys React + Flask. See the run script for individual gcloud commands you can use for any of the backends like /express, /ruby etc.
+This run script deploys React + Flask. See the run script for individual gcloud commands you can use for any of the backends like /express, /ruby etc. But if you're only deploying React, you need to use the full deploy.sh because it's running javascript build commands and sentry-cli.
```
cstavitsky, line 62: > But if you're only deploying React, you need to use the full deploy.sh because it's running javascript build commands and sentry-cli.

If you're only deploying react, do you need to comment out the flask deploy portion from `deploy.sh`? I feel this is getting overly complex for someone who doesn't know the ins and outs and have a suggestion.

How about separate `deploy_react.sh` and `deploy_flask.sh` scripts that are then stitched together in a single `deploy.sh`?

```
# deploy_react.sh

RELEASE=`./release.sh`
echo $RELEASE

SENTRY_ORG=testorg-az
SENTRY_PROJECT=application-monitoring-javascript
PREFIX=static/js

cd react
rm -rf build
npm install
npm run build

sentry-cli releases -o $SENTRY_ORG new -p $SENTRY_PROJECT $RELEASE
sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT set-commits --auto $RELEASE
sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT files $RELEASE upload-sourcemaps --url-prefix "~/static/js" --validate build/$PREFIX

# The release was set in the static prod build
gcloud app deploy --quiet
```

```
# deploy_flask.sh

cd ../flask && gcloud app deploy --quiet
```

```
# deploy.sh

source deploy_react.sh
source deploy_flask.sh
# possibly run through the whole list of other backends to deploy... or not
```

This way you can just run `deploy_react.sh` or `deploy_flask.sh` if you are only running one or the other. Other backends could also have their own deploy script i.e. `deploy_express.sh` or `deploy_springboot.sh`.

deploy.sh
```diff
@@ -14,7 +14,7 @@ sentry-cli releases -o $SENTRY_ORG new -p $SENTRY_PROJECT $RELEASE
 sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT set-commits --auto $RELEASE
 sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT files $RELEASE upload-sourcemaps --url-prefix "~/static/js" --validate build/$PREFIX
 
-# The release was set in the static prod build
+# This deploys React - The release was set in the static prod build
 gcloud app deploy --quiet
 
-cd ../flask && gcloud app deploy --quiet
+# cd ../flask && gcloud app deploy --quiet
```
cstavitsky, line 20: Is there a reason the flask deploy is commented out here?

express/db.js
```diff
@@ -135,9 +135,9 @@ function openDBConnection() {
     // public IP of the instance does.
     host = process.env.CLOUD_SQL_PUBLIC_IP
   } else {
-    host = process.env.CLOUD_SQL_CONNECTION_NAME
+    host = '/cloudsql/' + process.env.CLOUD_SQL_CONNECTION_NAME
```
cstavitsky, line 138: 👍

express/server.js
```diff
@@ -14,6 +14,20 @@
 
 'use strict';
 
+const makeRelease = function() {
```
cstavitsky, line 17: I suggest we name this `determineReleaseVersion` or something similar. When I read `makeRelease` my brain thinks that there's a release/deploy happening as a result of this function, whereas what it's really doing is determining the value of the release version.

express/server.js
```diff
@@ -60,7 +74,15 @@ Sentry.init({
     new Sentry.Integrations.Http({ tracing: true }),
     new Tracing.Integrations.Express({ app })
   ],
-  tracesSampleRate: 1.0
+  tracesSampleRate: 1.0,
+  tracesSampler: samplingContext => {
+    // sample out transactions for OPTIONS requests
+    if (samplingContext.request.method == 'OPTIONS') {
```
cstavitsky, line 80: When are `OPTIONS` requests being made within our app? Wasn't sure what situation led to this being added.

react/src/utils/backendrouter.js
```diff
@@ -1,16 +1,17 @@
...
+
 const SUPPORTED_BACKEND_TYPES = {
   "flask": {
       "test": "http://localhost:8080",
       "production": process.env.FLASK_BACKEND || process.env.REACT_APP_BACKEND
   },
   "express": {
       "test": "http://localhost:8088",
-      "production": process.env.EXPRESS_BACKEND
+      "production": process.env.REACT_APP_EXPRESS_BACKEND
```
cstavitsky, line 10: This file looks good. Note that we should update the `react/.env.example` to have `REACT_APP_` prefixes

https://github.com/sentry-demos/application-monitoring/blob/master/react/.env.example

i.e.

```
EXPRESS_BACKEND=
SPRINGBOOT_BACKEND=
FLASK_BACKEND
```

should all be updated to have `REACT_APP_` prefix.

react/src/utils/backendrouter.js
```diff
@@ -1,16 +1,17 @@
 const DEFAULT_BACKEND = "flask"
+
 const SUPPORTED_BACKEND_TYPES = {
   "flask": {
       "test": "http://localhost:8080",
       "production": process.env.FLASK_BACKEND || process.env.REACT_APP_BACKEND
```
cstavitsky, line 6: `FLASK_BACKEND` should be prefixed with `REACT_APP_` (or just deleted, alternatively, since we already have REACT_APP_BACKEND...)

spring-boot/src/main/appengine/app.yaml.example
```diff
@@ -1,13 +1,13 @@
 runtime: java
 env: flex
-service: application-monitoring-java
+service: <staging-springboot>
```
cstavitsky, line 3: this part is unclear -- does `<staging-springboot>` need to be replaced with something else in the actual app.yaml? could make it more explicit like, `<staging springboot service name goes here>`

README.md
thinkocapo: I'm expecting that if anyone wants to deploy, they should understand and appreciate how it's done (this also helps make sure they don't overwrite the Master instance) and will know to check the deploy.sh for relevant commands.

I'm not looking to add several more deploy.sh scripts right now, each for a different platform, especially if all they're doing is running a gcloud command. 

For now I've created a react/README.md (as the other platforms have README's) and put Deploy instructions there. The express/README and springboot/README's also have Deploy sections with instructions. I also added a react/deploy.sh since it's more than 1 command. I guess now there's gcloud commands for some platform deployments, and a deploy.sh for the react one. That's inconsistent but I want to get to he pytests, and adding more shell scripts can also confuse people - it'd still be a readme but pointing to a shell script instead of a gcloud command. And when the shell script doesn't work, they''ll have to understand gcloud and app.yaml's anyways. They really shouldn't be deploying if they don't understand it.

I'd rather not test a new deploy.sh that calls deploy_react.sh deploy_flask.sh and have to re-test all. But this will be helpful as you may need to deploy 1 or both or multiple services all at once. Happy to attempt it after I get the pytests running. @cstavitsky

deploy.sh
thinkocapo: this was a mistake, from when I was testing. But just now I added a react/deploy.sh so I can use that instead, and not have to bother editing this deploy.sh here

express/server.js
thinkocapo: These 'OPTIONS' http requests. I added the word 'http' if that makes it more clear.
Sometimes we open them during demo's by accident, and find that they're blank (there are no Spans inside of them). We have this in place in flask:
https://github.com/sentry-demos/application-monitoring/blob/master/flask/main.py#L40

This also allows us to have a valid example use case for tracesSampler ready in the app at all times, if we need to run an experiment for a customer.

https://sentry.io/organizations/testorg-az/performance/?project=6089805&statsPeriod=7d

The pre-flight 'OPTIONS' request to /products does not trigger the /products code to call the database (and hence no spans). It's not supposed to do that. It's just a pre-flight request, something related to CORS.
image

https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

@cstavitsky

react/src/utils/backendrouter.js
thinkocapo: good call

react/src/utils/backendrouter.js
thinkocapo: good call

spring-boot/src/main/appengine/app.yaml.example
thinkocapo: Hey so staging-applicaiton-monitoring-springboot was too many characters, it got rejected by Google when the gcloud command was run.

So I tried out staging-springboot because that reminded me of our discussion about maybe dropping the 'application-monitoring' part, as all of our App Engine services are for Application Monitoring (not sure it will always be that way).

Should I shorten them all? I could, in all the app.yaml.example files.

My concern with writing something like `<staging springboot service name goes here>` is that then they might follow no naming convention whatsoever. So I see `<staging-springboot>` and things like `<application-monitoring-flask>` as good placeholders.

I think the caret `<` also calls it to fail because it's not a valid character, which is good, because people should realize they might be overwriting the Master or Staging instances. If it's their first time applying, they should be updating the name here. 

I am not in support of the idea of , "I want to clone it....then run a deploy command a minute later...and not have to think about anything". I'm fine with the user going and researching what an app.yaml is about. We give enough instruction with, "Make sure your app.yaml has valid values", I don't want to write a bunch of instructions about every field.
@cstavitsky

express/server.js
thinkocapo: I renamed it.

README.md
cstavitsky: @thinkocapo understood, that's fine, thanks for explaining your perspective

express/server.js
cstavitsky: Thanks, was not familiar with the concept of Preflight requests, that answers my question

cstavitsky (approved):
👍

PR #49, opened 2021-12-08
SSL Cert error workaround documentation

cstavitsky:
- Document how to handle an SSL Cert error that prevents one from running tests locally
- Small refactor to make the local fix for this error easier to make (1 line local change instead of 6)


thinkocapo (approved):
looks good. go ahead and merge.

PR #50, opened 2021-12-09
Pull release artifacts from latest github release instead of hardcoded

cstavitsky:
- Previously the release number was hardcoded, and would have to be manually changed each time
- Going forward, the react native tests will no longer need to be updated with each new Github release


tests/release_version_manager.py
```diff
@@ -0,0 +1,5 @@
+import requests
+
+def latest_react_native_github_release():
+    react_native_releases = requests.get("https://api.github.com/repos/sentry-demos/sentry_react_native/releases")
+    return react_native_releases.json()[0]['tag_name']
```
thinkocapo, line 5: So this gets all of the releases and the 0th one is the most recent? Confirming. @cstavitsky

thinkocapo: ah duh the name of the method is latest...release, so that works! No need to answer me here.

tests/conftest.py
```diff
@@ -119,14 +120,15 @@ def android_emu_driver(request, data_center):
...
+    release_version = ReleaseVersion.latest_react_native_github_release()
 
     caps = {
         'username': username_cap,
         'accessKey': access_key_cap,
         'deviceName': 'Android GoogleAPI Emulator',
         'platformVersion': '10.0',
         'platformName': 'Android',
-        'app': "https://github.com/sentry-demos/sentry_react_native/releases/download/1.9/app-release.apk",
+        'app': f'https://github.com/sentry-demos/sentry_react_native/releases/download/{release_version}/app-release.apk',
```
thinkocapo, line 131: I didn't know Python 3 has these f'strings, as they're calling them. Good find.

It did trick me up at first, it reminded me of unicode strings in python like `ustring = u'A unicode \u018e string \xf1'` which is valid for writing in python.

thinkocapo (approved):
@cstavitsky how often are we calling api.github.com for the latest release? Is the saucelabs virtual device being re-created at a certain frequency?

No need to figure out the rate limit on the api.github.com api but if we ever run into problems reading from it, then it's something to think about.

Could use a try-catch and Sentry.capture_exception() if the call to react_native_releases fails, this way we could track it in our job monitor Project in Sentry that monitors TDA related problems.

Let's hold off, as there needs to be better instrumentation of handling failure scenarios in general across all of the tests (sentry-demos/tracing had a stronger implementation of this). Currently the test_add_to_cart is the only one with it.

As well as a sentry_sdk.capture_message somewhere indicating that a full suite of test (or one of the 5 pytests, is done in general). This is what I base the Alert on (it checks if the capture_message stops coming in at a certain frequency, it's how I know the job went down)

thinkocapo:
@cstavitsky the updated Release setting technique is approved.

PR #51, opened 2021-12-09
Pytest Platforms Express, Springboot, Enhancements

thinkocapo:
## Done
- The pytests `test_add_to_cart` and `test_add_to_cart_join` both select from flask, express, springboot backends at random
- Updated how the Express error is thrown in Checkout so that the stacktrace now shows the Inventory Code in it, instead of just the catch-block code. `throw new Error('Not enough Inventory')`
- **Fixed** 'cancelled' transactions by removing the 'wait statement in flask's /checkout process_order function.
- **Updated Express's** /checkout endpoint so it uses the **Automatically Instrumented transaction**, which fixes the broken distributed tracing (Manually Instrumented transactions breaks the reference to the traceId from React)
- Updated the springboot event's User object, by parsing the email from the incoming request's headers.
- Added Run script for running Springboot locally
- **React's Home Component was pulled out of index.js** and converted into a Smart Component in Home.js, so now it can make http calls to /success.
- The `test_homepage` pytest now calls backend /success routes so you don't see '????' in the Traceview.
image
and
image

## Testing
These are from the TDA job running locally from my computer, pointing to a mix of all 3 backends.

**Express**
[/checkout error](https://sentry.io/organizations/testorg-az/discover/will-express:48609f6723c14b00b8eaa1b419298d04/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=6089805&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)

[/products tx](https://sentry.io/organizations/testorg-az/discover/will-express:e1dffdc621f147e4a9278784dbbf0f15/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=6089805&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)

[/checkout tx](https://sentry.io/organizations/testorg-az/discover/will-express:822433db84f741b68dc3981e44c908fd/?field=title&field=event.type&field=project&field=http.url&field=timestamp&name=All+Events&project=6089805&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29) shows all 4 tx's + erro's linked via Distributed Tracing. It does not show the same spans as transactions in Flask, because spans must be manually finished `span.finish()` in Node, but they do not need that in python. In node, the error is thrown before the span can .finish(). In python, an error is thrown but the span is still able to .finish() on its own (because it's managed by the python sentry sdk, go look at the code). I do not expect parity yet between Express + Flask spans for the /checkout transaction. Solving that is beyond the scope of this PR.
image

**Springboot**
https://sentry.io/organizations/testorg-az/performance/trace/d7b3901d84274b6cade0f0c000529ee8/?pageEnd=2021-12-11T11%3A30%3A23.884&pageStart=2021-12-10T11%3A30%3A23.884
image

## TODO
make a query_string builder function for sharing this across tests


express/server.js
```diff
@@ -77,8 +77,9 @@ Sentry.init({
   ],
   tracesSampleRate: 1.0,
   tracesSampler: samplingContext => {
-    // sample out transactions for http OPTIONS requests
-    if (samplingContext.request.method == 'OPTIONS') {
+    // sample out transactions from http OPTIONS requests hitting endpoints
+    const request = samplingContext.request
+    if (request && request.method == 'OPTIONS') {
```
thinkocapo, line 82: sometimes there was not 'request' property on the samplingContext. I'm not quite sure why, but this fixed. Still getting the expected transactions coming in and nothing extra.

express/db.js
```diff
@@ -101,14 +101,26 @@ const getInventory = async function(cart) {
   }
   productIds = formatArray(productIds);
   console.log("> productIds", productIds);
+
   try {
-    let transaction = Sentry.startTransaction({ name: 'get inventory' });
-    let span = transaction.startChild({ op: 'get_inventory', description: 'db.query' });
+    const transaction = Sentry.getCurrentHub()
```
cstavitsky, line 106: Once again, great catch on this 👍

express/server.js
```diff
@@ -146,29 +149,37 @@ app.post('/checkout', async(req, res) => {
...
     console.log("> /checkout inventory", inventory);
-    span.finish();
-    transaction.finish();
 
+    span.finish();
+    
     // Process Order
-    transaction = Sentry.startTransaction({ name: 'process order' });
-    span = transaction.startChild({ op: 'process_order', description: 'function' });
+    let span2 = transaction.startChild({
```
cstavitsky, line 167: If you don't want to reuse the `span` variable as was previously being done, I recommend we name each span according to what they measure instead of using `span2`, which is not very descriptive.

i.e. `getInventorySpan` and `processOrderSpan` instead of `span` and `span2`.

tests/README.md
```diff
@@ -26,6 +26,8 @@ export SAUCE_ACCESS_KEY=<>
 chmod u+x tests/script.sh
 ```
 
+Make a .env and put DSN in their if you want catch errors for the pytests failing (not the Apps themselves having errors)
```
cstavitsky, line 29: typo `DSN in there`

tests/desktop_web/test_homepage.py
```diff
@@ -14,16 +16,19 @@ def test_homepage(desktop_web_driver):
     for endpoint in endpoints:
         sentry_sdk.set_tag("endpoint", endpoint)
 
-        # you can filter by se:tda in Sentry's UI
-        endpoint = endpoint + "?se=tda"
-
-        # Randomize the Failure Rate between 1% and 40%
-        n = random.uniform(0.01, .04)
+        for i in range(random.randrange(20)):
```
cstavitsky, line 19: Consider putting the query string generator (including failure rate) in its own function, as it's repeated 3 different times. Not totally necessary now but if we add even one more test that would have to copy the same code, could be useful to make it more reusable.

cstavitsky (approved):
Some small suggestions left in comments, overall looks good.

express/server.js
thinkocapo: The wordiness makes me lose track of the 'span' variables but they should be named. I might do spanGetInventory spanProcessOrder if that's okay. Helps me find it faster if I'm skimming the code. Maybe it's a personal preference for naming.

tests/desktop_web/test_homepage.py
thinkocapo: @cstavitsky I'd like to hold off on this re-factor only because the amount of re-testing (locally, staging, Master) I'd have to sit through is quite lengthy. I'm happy to do it on the next. I'm blocking someone else from some SDK upgrade PR's right now.

thinkocapo: I left a TODO in the code comments.

README.md
```diff
@@ -65,11 +65,14 @@ This run script deploys React + Flask. See the run script for individual gcloud
 ```
 and `gcloud auth login` if it asks you to authenticate, or insert YubiKey.
 
-#### How to Deploy to Staging
+## Deploy to Staging
```
thinkocapo, line 68: @sdzhong new section on how to deploy to Staging (or whatever you want to call it in your yaml's and .env file)

spring-boot/run.sh
```diff
@@ -0,0 +1 @@
+./mvnw spring-boot:run
```
thinkocapo, line 1: I added this run because the other platforms had it too and I had to keep checking the readme to copy/paste this mvnw command. @sdzhong

spring-boot/src/main/java/com/sentrydemos/springboot/AppController.java
```diff
@@ -65,7 +66,13 @@ private void setTags(HttpServletRequest request) {
 			String header = request.getHeader(tag);
 			if (header != null && header != "null") {
 				logger.info("Setting " + tag + " Sentry tag as " + header);
-				Sentry.setTag(tag, header);
+				if (tag == "email") {
```
thinkocapo, line 69: @sdzhong small update to set the email as the User info and no longer a tag.

flask/main.py
```diff
@@ -71,7 +71,6 @@ def checkout():
     print("> /checkout inventory", inventory)
 
     with sentry_sdk.start_span(op="process_order", description="function"):
-        wait(operator.ge, 14, .5)
```
thinkocapo, line 74: This 'wait was causing transactons+errors not to be linked to each other (Distributed Tracing), so removed it.

thinkocapo: @cstavitsky

PR #58, opened 2021-12-21
Update test to produce desired error for automated test data

cstavitsky:
- Previously, the react native android test was not producing errors. This PR updates that test so that it scrolls to the correct location and clicks the checkout button, triggering the 500 error we expect.


ndmanvar (approved):

thinkocapo (approved):
merge it when you're ready

PR #59, opened 2022-01-05
React Native Javascript Crashes in Homescreen Pytest

thinkocapo:
## Overview
This adds additional unhandled errors to React Native Android's TDA.

This lays the groundwork for Release Health in React Native.

'Crash Free Rate' is no longer stuck at '100%' as this new pytest triggers unhandled errors from button clicks on the home screen.

The randomization rate for erroring is set to 40%. Side Note: I realized there is no way to make it error 10% some days and 50% other days, because if you constantly pick a random number 1 to 10 over a 24hr period, it's always going to average out to 5 (think about it), which would give you 50%. So better just to pick a percentage, and I chose 40%. The code is simpler this way too.

## Testing
Run it 10 times in a row, and should not get 10 crashes, because it only crashes 40% of the time.
```
py.test -s -n 4 mobile_native/android_react_native/test_homescreen_react_native_android.py
[release](https://sentry.io/organizations/testorg-az/releases/?project=5716557)
```
Starting crashes count: 14
Finishing crashes count: Y

Start
image

Finish
image

## Next
1. Could try sending to background and back to foreground, which would create more sessions. This may aid our mobile metrics data too. 

"session is managed in the native layer (java/ios)...if you're doing that for testing/generating data, best choice is to lower the "sessionTimeout" option...to like 1 second or something, then just put the app on the background and on the foreground again" -Bruno.

2. Select the element by a more permanent attribute rather than a xpath, like:
```
driver.find_element_by_xpath('//*[@text="Unhandled Promise Rejection"]')
```
But, this was unable to find the element, perhaps because [this](https://chase-seibert.github.io/blog/2017/01/06/appium-react-native-quickstart.html). Note that you'd have to upload a new APK to Github to try this. It also has an example with setting up an accessibility_id.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -4,8 +4,8 @@ Runs automted tests against Sentry demos on GCP, in order to generate errors and
 ## Components / Moving parts
 - `conftest.py` -> Sauce Labs configuration (browsers) for frontend_tests
 - `backend_tests/backend_test.py` -> Hits /handled, /unhandled/, + /checkout backend demo APIs
-- Selectors for button elements in the React Native app can be found in Saucelabs via Sauce Labs Inspector...
-- [SauceLabs Inspector](https://github.com/appium/appium-inspector) download and launch this Desktop program, see img...TODO
+- Selectors for button elements in the React Native app can be found via connecting an Appium Inspector to your Saucelabs instance.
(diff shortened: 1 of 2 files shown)


cstavitsky (approved):
👍

PR #69, opened 2022-03-01
Springboot http client RestTemplate

thinkocapo:
## Overview
- **Springboot** - Upgraded sentry springboot sdk to 5.6.1
- **React** - renamed custom checkout transaction to 'Submit Checkout Form' which is what our react native demo app does now.
- **Flask** - fixed db error handling problem

HttpRestTemplate has been implemented in Springboot for making http requests.

Springboot **will soon** make http requests to Ruby. Currently the Ruby transaction gets made but is orphaned, it's not part of the trace with Javascript+Springboot, and so I commented out the ruby /api call in the springboot /products endpoint.

## Testing
Springboot [transaction](https://sentry.io/organizations/testorg-az/discover/results/?field=title&field=event.type&field=transaction.op&field=project&field=sdk.version&field=timestamp&name=All+Events&project=6090082&query=url%3A%22http%3A%2F%2Flocalhost%3A%2A%22&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29) with 5.6.1

[Submit Checkout Form](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:a1d0a7ff2d70464abbc1e418d81fc278/?field=title&field=event.type&field=transaction.op&field=project&field=user.display&field=timestamp&name=All+Events&query=se%3Awill&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29) transaction

The python error can only be replicated in prod.

### DIFF
--- a/flask/db.py
+++ b/flask/db.py
@@ -103,7 +103,6 @@ def get_products_join():
         raise DatabaseConnectionError('get_products_join')
     except Exception as err:
         err_string = str(err)
-        print(error_string)
         if UNPACK_FROM_ERROR in err_string:
             raise DatabaseConnectionError('get_products_join')
         else:
@@ -151,7 +150,11 @@ def get_inventory(cart):
     except BrokenPipeError as err:
         raise DatabaseConnectionError('get_inventory')
     except Exception as err:
-        raise(err)
+        err_string = str(err)
+        if UNPACK_FROM_ERROR in err_string:
+            raise DatabaseConnectionError('get_inventory')
+        else:
+            raise(err)
 
     return inventory
 
--- a/flask/db.py
+++ b/flask/db.py
@@ -75,7 +75,6 @@ def get_products():
         raise DatabaseConnectionError('get_products')
     except Exception as err:
         err_string = str(err)
-        print(error_string)
         if UNPACK_FROM_ERROR in err_string:
             raise DatabaseConnectionError('get_products')
         else:
--- a/flask/db.py
+++ b/flask/db.py
@@ -74,7 +74,9 @@ def get_products():
     except BrokenPipeError as err:
         raise DatabaseConnectionError('get_products')
     except Exception as err:
-        if UNPACK_FROM_ERROR in err:
+        err_string = str(err)
+        print(error_string)
+        if UNPACK_FROM_ERROR in err_string:
             raise DatabaseConnectionError('get_products')
         else:
             raise(err)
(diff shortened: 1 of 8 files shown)


flask/db.py
```diff
@@ -74,7 +74,8 @@ def get_products():
     except BrokenPipeError as err:
         raise DatabaseConnectionError('get_products')
     except Exception as err:
-        if UNPACK_FROM_ERROR in err:
+        err_string = str(err)
```
thinkocapo, line 77: stringifying the error because if you don't, then the next line errors with, "unable to iterate UNPACK_FROM_ERROR in err_string becasue err_string is an object and not a string"

the `err` exception is an object at first and not a string.

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
```diff
@@ -5,63 +5,63 @@
...
 	private static final Logger logger = LoggerFactory.getLogger(Application.class);
 	
 	public static void main(String[] args) {
 		SpringApplication.run(Application.class, args);
 	}
 
-	// sentry properties configured here for Java Spring Boot
-	// https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/	
 	@Bean
-	public void initSentry() {
```
thinkocapo, line 35: I was told to not use initSentry, as this is for regular java usage but not Springboot usage.
initSentry is in the regular java documentation but not Springboot's documentation.

This change was supposed to fix the trace-id getting onto the ruby transaction. It didn't.

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
```diff
@@ -5,63 +5,63 @@
...
-	private String getRelease() {
-		Date today = new Date(System.currentTimeMillis());
+	@Component
+	class CustomTracesSamplerCallback implements TracesSamplerCallback {
+	@Override
+		public Double sample(SamplingContext context) {
 
-		logger.info("release: " + today.getMonth() + "." + ((today.getDay() - 1)/7+1));
-		return today.getMonth() + "." + (today.getDay() - 1/7+1);
+			CustomSamplingContext customSamplingContext = context.getCustomSamplingContext();
```
thinkocapo, line 46: I'm still working on getting this to filter out OPTIONS requests. It didn't quite work on first attempts, so leaving it for future development.

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
```diff
@@ -5,63 +5,63 @@
...
+				return 1.0;
+			}
+		}
 	}
-	
-}
\ No newline at end of file
+}
+
+// Sentry.init...
```
thinkocapo, line 57: The old way we used to sample out OPTIONS requests, leaving it here for now.

ndmanvar (approved):

PR #72, opened 2022-05-05
Updated RELEASE val with package

sdzhong:
- [x] Bugfix
 - [ ] New feature
 - [ ] Enhancement
 - [ ] Refactoring

### Description
Local Empower Plant frontend source maps are not getting used which manifested in error `Cannot fetch resource due to restricted IP address`

The change is updating ./run.sh to use package name as a part of `RELEASE`; [matching react/src/index.js](https://github.com/sentry-demos/application-monitoring/blob/master/react/src/index.js#L49).

### Before
Before this change, [an Issue event's source maps are not applied](https://sentry.io/organizations/testorg-az/issues/3261093431/events/a5320aee02364a03bd1d5d09d7347b53/?project=5952962) and multiple releases are created (e.g. application.monitoring.javascript@2.2.2 and 2.2.2).
image

### After
By applying this change, [source maps are applied as expected](https://sentry.io/organizations/testorg-az/issues/3261078026/events/533ff2f1ba714afab514e7947a48ffe3/?project=5952962) and the single correct release is created (e.g. application.monitoring.javascript@2.2.3).
image

### DIFF
--- a/react/run.sh
+++ b/react/run.sh
@@ -1,6 +1,7 @@
 # Builds and serves the js bundle, uploads sourcemaps and does suspect commits
-
-RELEASE=`../release.sh`
+PACKAGE="application.monitoring.javascript"
+VERSION=`../release.sh`
+RELEASE=$PACKAGE@$VERSION
 echo $RELEASE
 
 SENTRY_ORG=testorg-az
--- a/react/run.sh
+++ b/react/run.sh
@@ -1,6 +1,6 @@
 # Builds and serves the js bundle, uploads sourcemaps and does suspect commits
 
-PACKAGE="application.monitoring.javascript"
+PACKAGE=application.monitoring.javascript
 VERSION=`../release.sh`
 RELEASE=$PACKAGE@$VERSION
 echo $RELEASE
--- a/react/run.sh
+++ b/react/run.sh
@@ -1,4 +1,5 @@
 # Builds and serves the js bundle, uploads sourcemaps and does suspect commits
+
 PACKAGE="application.monitoring.javascript"
 VERSION=`../release.sh`
 RELEASE=$PACKAGE@$VERSION


thinkocapo:
@sdzhong I've given several demos lately and haven't seen this is a problem. The issue you linked is for the 'test'environment, see its tag. 

Could you deploy this to staging and see if it works? 

Don't want to risk breaking anything in production.

thinkocapo:
@sdzhong I think these changes would have to go in deploy.sh because that's what staging and production use. I haven't opened your issue link but my guess is you're running run.sh locally. Do you have the right Project and Org values in your run.sh? Are you checking the right Project that corresponds to it? Looks like you're checking Simon-react , according to your issue, and i think run.sh is uploading to a different project, unless you reverted the change for the PR.

Src/index points to application-monitoring-react so you could use that instead of simon-react, and add se=simon to the query string, and filter by that in Discover, you can filter by url=localhost... too. This way you never have to mess with values in run.sh, deploy.sh, .env or src/index.js, to do local runs

I'm not sure where src/index is getting the package name from, it could vary depending is it testing or prod being used. You're welcome to write simon.react there if you'd like. 

Maybe sentry-cli doesn't require semantic versioning in the release name. Could check documentation or ask. Either way I'd deploy to staging and make sure it works. I can review next week.

sdzhong:
Yes, this problem appears when running the react app locally (e.g. localhost:5000), not on the public demo.

@thinkocapo, would you like me to:
1. Deploy to Staging as is, then validate?
2. Make corresponding changes to `react/deploy.sh`, deploy to staging, then validate?

thinkocapo (approved):
looks good, merge it. Then I'll deploy it to prod and monitor it. Thanks for talking this one through. Nice work.

PR #75, opened 2022-06-07
this is a test to show Prithvi PR stuff - [from fork] Demonstration to show Prithvi forks

cstavitsky:
......

Screen Shot 2022-06-07 at 11 32 16 AM

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,7 +22,7 @@ print("DSN", DSN)
 
 import urllib3
 urllib3.disable_warnings()
-
+# Initialize sentry sdk
 sentry_sdk.init(
     dsn= DSN,
     traces_sample_rate=0,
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1,3 +1,4 @@
+# Temp comment
```
cstavitsky, line 1: Not sure about this change, have you considered doing X instead?

cstavitsky, line 1: Comment 1

tests/conftest.py
```diff
@@ -1,3 +1,4 @@
...
 import pytest
 from os import environ
 import os
```
cstavitsky, line 4: Comment 2

cstavitsky (commented):
Had a few questions before approving

PR #76, opened 2022-06-09
Crash Free Rate Adjustment

thinkocapo:
## Overview
Today, all releases have a Crash Free Rate of ~90%. We'd like to alternate the crash free rate each week between 90 and 95%.

Currently the crash rate is set at 1 to 40%, on the test that hits the home page. So home page crashes 1 to 40% of the time. The Checkout workflow is not controlled by a crash rate. It always crashes. So it's the Home Page that really affects the Crash Free Rate.

So let's try 1 to 40% and then on the alternate week do 1 to 20% so that we'll get higher than 90%.

Let it run a few weeks and then we can adjust.

## Testing
Would really need to let this run for a full 2 weeks, but here's a script to test it calculates the right week number and a value based on that.

It's week 2 today which means even, so the rate should always be 1 to 20%, as opposed to 1 to 40%. So it maxes out at 0.2

```
 ~/tmp   cat week.py
from datetime import datetime
import random

d=datetime.today()
week=str((d.day-1)//7+1)
week=int(week)

print(week)

# For setting a different Crash Free Rate each week
upper_bound = 0

if week % 2 == 0:
    # even patch version, e.g. 22.6.2
    upper_bound = .2
else:
    # odd patch version, e.g. 22.6.3
    upper_bound = .4

n = random.uniform(0.01, upper_bound)

print(n)
 ~/tmp   python3 week.py
2
0.13836519611835457
 ~/tmp   python3 week.py
2
0.036610909266023485
 ~/tmp   python3 week.py
2
0.17158809796399538
 ~/tmp   python3 week.py
2
0.1386184034128947
 ~/tmp   python3 week.py
2
0.06574793037712336
 ~/tmp   python3 week.py
2
0.055573418897122105
 ~/tmp   python3 week.py
2
0.0935466549573589
```


cstavitsky (approved):
👍

PR #77, opened 2022-06-12
Added a vue project to the mix

Prithvirajkumar:
This PR is to append our vue project to our application-monitoring project. It is currently maintained outside of our sentry-demos org and the goal is to maintain and manage the repo here.


Prithvirajkumar:
the vue project has an existing git file and that has caused some issues so I am closing this PR.

PR #79, opened 2022-06-21
Update selectors for android native tests

cstavitsky:
- Update selectors for android native tests, now that app launches into empower plant directly
- Not updating these would break our automated demo data for android


thinkocapo (changes requested):
@cstavitsky could you put a link to a Trace from either the main screen (EmpowerPlant/Home) startup which has products, or from Checkout, so we can see the result? Testing. 1 link is fine.

thinkocapo (approved):

PR #82, opened 2022-07-05
React Refactor - More Components Profiled

thinkocapo:
## Overview
I refactored the Products and ProductsJoin components by compoentizing the Product card in the UI that loads for each product. So now you'll see `<Products>` span as well as `<ProductCard>` for each product. Previously you did not see 4 `<ProductCard>` spans. This was unrealistic, because enterprise web apps have large component hierarchies and would use a smart component for such a things (see Issues stream component hierarchy in getsentry, for example).

This shows off React's Profiler better, now that each Product/:id item has its react component profiled and shown in the transaction.

Now it's easier to point out, "The slow /products call, blocked each **Product Card Component** from rendering [show the span and ui.react.mount, ui.react.update spans]".

The `SET_FLAG` Action+Reducer was added so that the Product/:id component gets its `props` updated, and this is what causes the `ui.react.update` span to appear. Without updating `props` in the component, you would only get ui.react.mount but not ui.react.update.

Future - this allows us to demonstrate sequential vs parallel performance problem via web architecture, now that we can point out using a full Component span, rather than just a http span. Not all users of Performance Monitoring are installing on backend, so sequential vs parallel should not only exist in backend.

Ignore the branch name, no actual slowness was introduced, though this arguably makes it seem slower now you can see more inner workings being held up by the slow /products endpoint 😎 

## Testing
[/products transaction.pageload](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:5f942ed5d69b4069b9b9c8bfa4841c09/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=se%3Awill+project%3Aapplication-monitoring-javascript&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)

[/products-join transaction.pageload](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:4f66218b21b040eba7737ea79133fd2f/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=se%3Awill+project%3Aapplication-monitoring-javascript&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)

[/checkout trace](https://sentry.io/organizations/testorg-az/performance/trace/eaca6042d7df4773a57091fd5f6679e3/?pageEnd=2022-07-07T01%3A07%3A33.426&pageStart=2022-07-06T01%3A07%3A33.426) still looks fine

image

and redux-logger shows the state of the Flag being accurately toggled, which updates props, so that we get the ui.react.update span
image

## Next
Option - sequential-vs-parallel. Initial fetch /products without reviews data. Empower Plant adds reviews microservices, so then do a separate /product/:id (needs to be created in backend) http request for the reviews data. This creates a sequential problem, the Problem State. Then show an After State, the Fixed State, which uses parallelization by gathering this data initially in the /products request (and does not rely on /product/:id anymore).


react/src/components/ProductsTree.js
```diff
@@ -0,0 +1,121 @@
+// import { Component } from 'react';
```
cstavitsky, line 1: If we're adding a new file that's entirely commented out, consider adding a comment at the top explaining under what conditions we would comment it back in? Or if you plan on re-commenting soon, no problem and no need to do anything

cstavitsky (approved):
LGTM, left a very small comment

thinkocapo:
Noting that I updated Staging instances for Express and Springboot so now staging calls to /api /connect /organization work. They were already working in Production

[/products trace with Express backend](https://sentry.io/organizations/testorg-az/performance/trace/cf08333d5b8f430c8b954ba60797e119/?pageEnd=2022-07-08T21%3A35%3A40.486&pageStart=2022-07-07T21%3A35%3A32.618)

[/products trace with Springboot backend](https://sentry.io/organizations/testorg-az/performance/trace/12b13639ae404fe3ae59736db5b42beb/?pageEnd=2022-07-08T21%3A35%3A26.192&pageStart=2022-07-07T21%3A35%3A17.895)

[/products trace with Flask backend](https://sentry.io/organizations/testorg-az/performance/trace/61596df7dc0e46e99241a4e380885340/?pageEnd=2022-07-08T21%3A34%3A20.397&pageStart=2022-07-07T21%3A34%3A12.898)

@cstavitsky

PR #83, opened 2022-07-15
Add: filter out OPTIONS and limit email addresses. 

bfpepper87:
Type of change: 

- [ ] Bugfix
- [x] New Feature
- [ ] Enhancement
- [x] Refactoring

Description: 

Added the ability to filter out OPTIONS requests on SpringBoot. This allows a user to NOT consume quota for requests that don't have any real value. Can change the request type to suit a demo's need. Added CaptureConsoleIntegration because it's a feature we were missing

Added better email tag generation. This will limit the number of unique emails that get passed to the UI. This will allow the UI to not have a 1:1 match for Issues and Users.


react/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@sentry/integrations": "^7.2.0",
```
thinkocapo, line 6: Hey what prompted this and the `CaptureConsoleIntegration` in index.js?

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
```diff
@@ -44,24 +44,15 @@ class CustomTracesSamplerCallback implements TracesSamplerCallback {
 		public Double sample(SamplingContext context) {
 
 			CustomSamplingContext customSamplingContext = context.getCustomSamplingContext();
-			if (customSamplingContext != null) {
```
thinkocapo, line 47: are you sure we should remove this null check here? What is is null one day? Then I think the following line would error:
```
HttpServletRequest request = (HttpServletRequest) customSamplingContext.get("request");
```

bfpepper87:
Discovery showing that on the 14th, OPTIONS requests stopped being processed. 

https://sentry.io/organizations/testorg-az/discover/results/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=6564946&query=&sort=-timestamp&statsPeriod=7d&yAxis=count%28%29

ruby/main.rb
```diff
@@ -12,6 +12,8 @@
 # This must be imported after sentry-ruby for transactions to work
 require "sinatra"
 require "sinatra/cors"
+require "google-cloud-datastore"
```
thinkocapo, line 15: @bfpepper87 let's get rid of this if this is just from testing.

ruby/main.rb
```diff
@@ -12,6 +12,8 @@
...
+require "sinatra/activerecord"
```
thinkocapo, line 16: @bfpepper87 let's get rid of this if it's not being used. Please make sure the Gemfile didn't get updated with this too

ruby/main.rb
```diff
@@ -30,6 +32,10 @@
   "Sentry Ruby Service says Hello - turn me into a microservice that powers Invoicing, Trucking, or DriverFind"
 end
 
+# get"/products" do
```
thinkocapo, line 35: @bfpepper87 let's get rid of this since I think we didn't finish developing this. Maybe you should switch git branches if you're switching to a new demo feature, so you don't confuse code. OPTIONS demo task is different than the ruby stuff you were working on.

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
thinkocapo: if customSamplingContext is null/undefined then it will say "can't read .get of undefined".

java likes to do a lot of null checks, and there's a good chance whatever boilerplate code I got that from a while ago, it probably gave me that null check. I tend to not insert them myself.

thinkocapo: sounds good.

spring-boot/src/main/java/com/sentrydemos/springboot/Application.java
```diff
@@ -44,24 +44,15 @@ class CustomTracesSamplerCallback implements TracesSamplerCallback {
...
-				HttpServletRequest request = (HttpServletRequest) customSamplingContext.get("request");
-				return 1.0;
+
+		HttpServletRequest request = (HttpServletRequest) customSamplingContext.get("request");
```
thinkocapo, line 48: @bfpepper87 I just realized that line 48 will fail if the null check is not performed first. I see you moved the null check to down below to line 50, but I checked the current `master` code and it looks like the null check should be executed before your line 48 as well as before line 50. Could you re-arrange here? Or what do you think?

image

thinkocapo: the purpose of the null check is that so the following line won't fail to finish executing:
`HttpServletRequest request = (HttpServletRequest) customSamplingContext.get("request");`
With your most recent code, the null check occurs after that line, not before it (like in master). Unless I'm reading it wrong.

thinkocapo (approved):

PR #84, opened 2022-07-20
FE React Constructor delay, BE delay adjustment

thinkocapo:
## Overview

[transaction link](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:0722e179b5714c458c67a927b5914815/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=se%3Awill&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)

Story,
"There's some kind of delay going on in this react component mounting"
"Which is due to the `constructor` function, so maybe something async in there"
"maybe need to leverage component callback functions better, or maybe not and there's a legitimate delay in whatever's being performed..."

adjust /products Wait/weighter function.

Note - realized there's no weight distribution on the reviews sql query, it's always 0.0625.

Note - constructor delay is only on Products.js and not ProductsJoin. This is be design because we want ProductsJoin to be fast.

## Testing
[Discover link](https://sentry.io/organizations/testorg-az/discover/results/?end=2022-08-12T11%3A35%3A00&field=project&field=title&field=transaction.duration&field=delay&field=se&field=timestamp&name=All+Events&project=5808655&query=event.type%3Atransaction+AND+%28url%3Ahttp%3A%2F%2Flocalhost%3A%2A+OR+se%3Awill%29+AND+title%3Aproducts&sort=-timestamp&start=2022-08-12T10%3A35%3A04&yAxis=count%28%29)

Before, 5 times
times = [0.05, 0.25, 0.5, .75, 1.0, 1.25]
image

After, 5 times - reduced delays 50%. it made them a little bit shorter but not much
times = [0.025, 0.125, 0.25, 0.375, 0.5, 0.625]
image

After - reduced delays another 50%. it made a tiny bit of a difference again? is this worth it?
image

I'm not sure why the query is taking 7 to 11 seconds, based on the values we're passing to `pg_sleep` here:
```
"SELECT *, pg_sleep(%s) FROM products" % (n)
```
some [pg_sleep documentation](https://database.guide/how-pg_sleep-works-in-postgresql/)

so trying...

After - new set of delay values. it seems passing whole integers do not correlate 1:1 to whole seconds
times = [3, 4, 3, 2, 1, 1]
image

UPDATE - I realize the delay is coming from the SQL query that selects the reviews, where the delay is hard-coded and does not select from a distribution via the random.choices and weighter() functions.

After - with reviews delay set to 0.125 (50% reduction from 0.25), but back to regular delay time on the first products query. much shorter finally:
image

After - with reviews delay set to 0.125 (50% reduction from 0.25), and products delay at 50% reduction
image

After - with reviews delay set to 0.0625, and products delay still at 50% reduction
image

After - with reviews delay set to 0.0625, and products delay at 2x 50% reduction
image


cstavitsky:
Since the span (screenshotted) looks about the same length as the HTTP request, it doesn't necessarily immediately scream 'slow', would we want to increase the slowdown time at all? I think it's OK though. 

One high level thought I have more generally (would have mentioned earlier if i'd thought of it) is that it might be good to have a frontend slowdown be one that is custom instrumented (language/SDK agnostic). My reasoning is that if we want our demos to be consistent, it may be a little harder to find the equivalent of a `React.mount` in mobile apps, and we end up with our demos starting to get fragmented. For instance if someone doesn't use React, is a `React.mount` something that they will be impressed by? Whereas if we show something broader like, "there was a slowdown in the flow of a user adding a product to their cart" then that can be done on mobile or any other non-react frontend.

Screen Shot 2022-07-20 at 9 57 17 AM

thinkocapo:
@cstavitsky the http request fluctuates between like 2seconds and 10seconds. I don't think it needs to be wider than the http request. I'm not sure what would make it scream 'slow anyways. I don't see it as needing to be longer. If we have too many wide spans then it could skew anything anyways or start to make it look like it's normal to have wide spans.

I like the idea of **language/SDK agnostic** problem but I don't see a good way to do it, since in JS you're running in a browser (which is basically an app on top of the user's Operating System) where as in mobile you're running directly on the Operating System, have access to file I/O, database, hard drive, threads, and a Browser only has like a file upload API there's not much else it can do to its host operating system) so I think realistic slowness problems are fundamentally different. Trying to force fit one for all platforms could take away from the best examples we could actually be showing (e.g. threading/database problems in mobile, like we'll demo in Profiling soon). But thanks for raising the idea. Unless @ndmanvar thinks otherwise.

We wanted something React specific in this case.

ndmanvar (approved):

PR #85, opened 2022-07-27
Fix: Read headers in ruby

sergiosentry:
Type of Change:

- [x]  Bugfix
- [ ]  New feature
- [ ]  Enhancement
- [ ]  Refactoring

Summary:

Ruby backend was not parsing the headers sent, so it was missing information like the `se`, `customerType` and `email` tags. Since the flask backend is also making a call to the ruby backend, I added the headers there as well.

Transactions

[link](https://sentry.io/share/issue/c019bc3c7fa6423f95e2ca699a5dcfac/)

### DIFF
--- a/flask/main.py
+++ b/flask/main.py
@@ -7,7 +7,7 @@ from flask import Flask, json, request, make_response
 from flask_cors import CORS
 from dotenv import load_dotenv
 from db import get_products, get_products_join, get_inventory
-from utils import release, wait
+from utils import release, wait, parseHeaders
 import sentry_sdk
 from sentry_sdk.integrations.flask import FlaskIntegration
 from sentry_sdk.integrations.sqlalchemy import SqlalchemyIntegration
@@ -17,6 +17,7 @@ RELEASE = os.environ.get("RELEASE") or release()
 DSN = os.getenv("FLASK_APP_DSN")
 ENVIRONMENT = os.environ.get("FLASK_ENV") or "production"
 RUBY_BACKEND = os.environ.get("RUBY_BACKEND")
+RUBY_CUSTOM_HEADERS = ['se', 'customerType', 'email']
 
 print("> DSN", DSN)
 print("> RELEASE", RELEASE)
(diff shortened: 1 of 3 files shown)


flask/main.py
```diff
@@ -97,15 +97,15 @@ def products():
         raise(err)
 
     try:
-        r = requests.get(RUBY_BACKEND + "/api")
+        r = requests.get(RUBY_BACKEND + "/api", headers=request.headers)
```
thinkocapo, line 100: @sergiosentry let me think this through. I'm concerned if we re-use the same headers, then we'll lose any headers that the requests.get request was going to set itself, as well as any request headers that App Engine (the production environment) will be adding. So I'd at least need to test this in staging first.

Basically the request headers on the Ruby transaction would not longer be the real request headers sent by Flask. Instead, they'd be the request headers sent from React-to-Flask. Does that make sense?

It might be better to set se, customerType and email one-by-one like:
```
headers={se, customerType, email}
```
and then I think all the other headers by the requests library and AppEngine would still get set. I'm just concerned we'd be overriding them all if we did `headers=request.headers`. Let me get some opinions on this.

thinkocapo: image

flask/main.py
```diff
@@ -114,7 +119,12 @@ def products_join():
         raise(err)
 
     try:
-        r = requests.get(RUBY_BACKEND + "/api")
+        headers = {
+            "se" : None if request.headers.get('Se') == 'undefined' else request.headers.get('Se'),
+            "customerType" : None if request.headers.get('Customertype') == 'undefined' else request.headers.get('Customertype'),
+            "email" : None if request.headers.get('Email') == 'undefined' else request.headers.get('Email')
+        }
+        r = requests.get(RUBY_BACKEND + "/api", headers=headers)
```
sergiosentry, line 127: Alternatively, we could add a method on `utils.py` that parses the headers and call it from here, since this same code is repeated twice.

thinkocapo: @sergiosentry sounds good, yes please do that with `urils.py`

One question.

How did you settle on checking the string as `.get('Customertype')` versus `.get('customerType')`? We try to write it as customerType, coming from the client side (web JS, mobile), and then check for it as 'customerType' in our server api (here's [how Flask does it](https://github.com/sentry-demos/application-monitoring/blob/a125ed8dd3a448c949b0e90e506434ba9c48c119/flask/main.py#L156), as an example), but maybe ruby force lowercasing it then capitalizing the first letter? Is that what you found?

I guess it's possible that Express, Spring and Flask all force lowercase it as they send the request outbound to Ruby? I feel like I've seen the casing of request header strings get adjusted before, bot not sure.

Did you re-test? Could you update the transaction link showing that it's working for React<>Flask<>Ruby?

sergiosentry: @thinkocapo Got it!
Mhmm weird, I can get the values of the headers by using both `customerType` and `Customertype`. I guess is not case-sensitive for the `request.headers` object in Flask? I do know that Ruby converts headers to uppercase by convention [reference](https://jatindhankhar.in/blog/custom-http-header-and-ruby-standard-library/#:~:text=Since%20for%20most%20of%20the,of%20the%20third%20party%20apis.) 

Anyways, I changed it to use the same casing as the react frontend (`customerType`) and it works! I also updated the transaction link

thinkocapo (approved):

PR #90, opened 2022-08-10
IGNORE - FE Performance Delay

thinkocapo:
IGNORE THIS

See https://github.com/sentry-demos/application-monitoring/pull/84/files instead


thinkocapo:
Consider putting these changes on https://github.com/sentry-demos/application-monitoring/pull/84

PR #91, opened 2022-08-11
SDK Upgrades and React Router v6 for Dynamic Sampling

thinkocapo:
## Dynamic Sampling
Upgraded sdk's to
javascript.react 7.10.0
javascript.node 7.11.1
python 1.9.3

#### Note on Testing
In order for Dynamic Sampling to be configurable in the UI and not greyed out, the latest SDK versions must be present, and this means your event data must be at 5% of adoption. So we won't reach 5% unless we deploy the updated SDK's to production, because TDA is running on that 24/7. If you only run the updated SDK's locally, then the amount of times you manually load these apps will never be 5% of all the data TDA is producing, and so you'll never be able to turn on Dynamic Sampling.

## Why Dynamic Sampling Failed The First Time
Implementing a better react-router configuration in general is the solution to preventing the `<<unparameterized transactions>>` grouping of all our transactions. In the process of implementing this, I figured it was better to use the time to upgrade our version of react-router too, for several reasons.

It was pointed out we weren’t passing enough params into our Sentry.init(integrations: Sentry.ReactRouter(param1,param2,param3). We only had param1, I believe this was the minimum config when it was first developed. However passing the param2, param3 was going to cause a lot of clunky code (see v5 documentation). v6 was a lot cleaner and more developers are using that too today. We’d rather debug problems off v6 as well

## React-Router v5->v6 - Why
React Router v6 is more widely used today. It also has a cleaner code implementation (you write `<SentryRoutes>` once and wrap routes with it, rather than wrap each individual Route with a `<SentryRoute>`). Read about the differences in config [here](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/). We'll get better support from Sentry developers if we're on the more updated version of react-router

## React-Router v5->v6 - What Changed
So we upgraded to it, and it required we turn some Smart Components into Functional Components so they would work with React Router v6. This was because the code we used for things like passing params and performing redirects,  were dependent on the version of react-router. In order for these to still work, we had to update them to how react-routerv6 requires them be done.

## Class Components to Functional Components
Noting that there are still some Class Components in the `/components` directory. This is okay, as now we have a mix of both and customers will be on a mix of both. Nothing inherently wrong with having both. Spending hours refactoring the rest does not have a visible effect in a demo nor provide value to the customer in a demo, so not bothering refactoring the rest at this time.

## Some Changes Made
- Employee.js, Product.js, Checkout.js Smart Components turned into Functional Components.
- useParams now used for parsing URL Params, per documentation v6 https://reactrouter.com/en/v6.3.0/getting-started/overview#configuring-routes
- no more Route Switch needed
- Redux and Context still used in some Function Components.
- useState for local state within Function Components
- useEffect for fetch'ing resources in Function Components.

## Testing
Make sure Home, About, Products->Checkout workflows still generate the same events.

This PR is not turning on Dynamic Sampling and testing. This PR is only putting the right code into place so that DS can be tested in the future.


ruby/main.rb
```diff
@@ -26,8 +27,8 @@
 require "sinatra/cors"
 
 set :allow_origin, "*"
-set :allow_methods, "GET,HEAD,POST"
-set :allow_headers, "content-type,if-modified-since,accept,access-control-request-headers,access-control-request-method,origin,sec-fetch-mode,user-agent,customerType,email,Referer,se,sec-ch-ua,sec-ch-ua-mobile,sec-ch-ua-platform,sentry-trace"
+set :allow_methods, "GET,HEAD,POST,OPTIONS"
+set :allow_headers, "baggage,content-type,if-modified-since,accept,access-control-request-headers,access-control-request-method,origin,sec-fetch-mode,user-agent,customerType,email,Referer,se,sec-ch-ua,sec-ch-ua-mobile,sec-ch-ua-platform,sentry-trace"
```
thinkocapo, line 31: for propagating the sampling context. the javascript sdk v7.X.X now includes this header.

ndmanvar (approved):

thinkocapo:
changing 'staging' to production before merging. @cstavitsky

thinkocapo:
This PullRequest is on hold because upon deploying, we found that auto-instrumented transactions were getting grouped together as `<<unparameterized transactions>>`. These were not manual transactions or high-cardinality transactions. These were auto-instrumented transactions.

So we reverted the JS SDK back to 6.10.1 and now the data looks good again. We're also turning off the Dynamic Sampling feature flag from our demo org for the time being.

thinkocapo:
All of this got deployed to environment:staging with updated SDK's, but we reverted the v7.X.X Javascript SDK back to v6.X.X because of the `<< unparameterized transactions >>` problem. So ruby flask springboot express still have updated SDK's, but only for staging.

We've since turned off the feature flag for Dynamic Sampling for our demo test org.

So do not merge this yet.

dgbailey:
@thinkocapo was this resolved https://github.com/sentry-demos/application-monitoring/pull/91#issuecomment-1227245467
? Seems like we are moving forward with the latest major for JS sdk but I may have missed the writeup on this.

thinkocapo:
@dgbailey yes moving forward with major 7.10.0 for JS SDK. 7.10.0 was already tested. All is resolved now.

Originally we had 7.10.0 then I said to Hold Off because of a problem, then we realized we needed react-router update and better configuration. So then added that in. I know it's fixed for 7.10.0, don't want anything higher because we've only tested 7.10.0 and don't want to create new bugs just yet.

dgbailey (approved):

PR #92, opened 2022-09-06
[IGNORE] Dynamic Sampling - React Router v5 and SentryRoute Upgrade

thinkocapo:
### React Router sentry documentation
https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/#react-router-v6

https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/#react-router-v4v5

### How to Pass Props to SentryRoute
```
<SentryRoute path="/product/:id" render={routeProps => (
                    <Product {...routeProps} backend="myvalue" />
                )}></SentryRoute>
```

If we switch to React Router v6, we may lose purpose in using beforeNavigate, as it would now get grouped+parameterized correctly via `<SentryRoutes>` wrapping all `<Route>'s`.

### DIFF
--- a/react/src/components/About.js
+++ b/react/src/components/About.js
@@ -16,7 +16,7 @@ const employees = [Jane, Lily, Keith, Mason, Emma, Noah];
 class About extends Component {
 
   async shouldComponentUpdate() {
-    console.log("> About shouldComponentUpdate")
+    // console.log("> About shouldComponentUpdate")
   }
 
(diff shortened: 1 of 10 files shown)


thinkocapo:
Ignore this because we prefer React Router v6 which is being done in https://github.com/sentry-demos/application-monitoring/pull/91

PR #93, opened 2022-09-13
Update deploy.sh

quickstark:
Updated deploy.sh to Finalize Release and add production tag.

[Release finalization ](https://docs.sentry.io/product/cli/releases/#finalizing-releases) is intended to explicitly define what is considered as the "next release" (for purposes of fixing issues) as well as  aligning commits.  No evidence finalizing a release impacts sentry.io UI's sort order in the Release tab (see GIF)

Sentry Release Finalization

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,5 +1,6 @@
 RELEASE=`./release.sh`
 PACKAGE_NAME=application.monitoring.javascript
+ENVIRONMENT=production
 
 RELEASE=$PACKAGE_NAME@$RELEASE
 echo $RELEASE
@@ -14,8 +15,10 @@ npm install
 npm run build
 
 sentry-cli releases -o $SENTRY_ORG new -p $SENTRY_PROJECT $RELEASE
+sentry-cli releases -o $SENTRY_ORG finalize -p $SENTRY_PROJECT $RELEASE
 sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT set-commits --auto $RELEASE
 sentry-cli releases -o $SENTRY_ORG -p $SENTRY_PROJECT files $RELEASE upload-sourcemaps --url-prefix "~/static/js" --validate build/$PREFIX
+sentry-cli deploys -o $SENTRY_ORG new -p $SENTRY_PROJECT -r $RELEASE -e $ENVIRONMENT -n $ENVIRONMENT
 
 # This deploys React - The release was set in the static prod build
 gcloud app deploy --quiet


deploy.sh
```diff
@@ -14,8 +15,10 @@ npm install
 npm run build
 
 sentry-cli releases -o $SENTRY_ORG new -p $SENTRY_PROJECT $RELEASE
+sentry-cli releases -o $SENTRY_ORG finalize -p $SENTRY_PROJECT $RELEASE
```
realkosty, line 18: does it matter if you call finalize after `set-commits` or after?

thinkocapo: @ndmanvar any thoughts here? Can we get your review on this PR?

deploy.sh
```diff
@@ -1,5 +1,6 @@
 RELEASE=`./release.sh`
 PACKAGE_NAME=application.monitoring.javascript
+ENVIRONMENT=production
```
thinkocapo, line 3: Noting this is fine, because the only time you set it as 'production' is when you deploy to app engine, which is what deploy.sh is for. So it matches with what is in src/index.js, should be no conflicts where sentry-cli is setting one value and src/index.js is setting a different one:

```
let ENVIRONMENT
if (window.location.hostname === "localhost") {
  ENVIRONMENT = "test"
} else { // App Engine
  ENVIRONMENT = "production"
}

...
Sentry.init({
  dsn: DSN,
  release: RELEASE,
  environment: ENVIRONMENT
...
```

thinkocapo: ^ but interesting this is not pulling from the scripts defined in package.json, I suppose. It totally could, but `npm run build` is used for testing locally, as well as for uploading a build with the deploy script, therefore you'd sometimes want it to say 'test' (when running locally) and sometimes want it to say 'production' (when deploying), so I found it more convenient to use the window.location.hostname as the indicator to whether you're running in Test or Production.

The run.sh script is how you run locally, it uses npm run build (defined in package.json). Or run npm run start, manually.

ndmanvar (approved):

deploy.sh
realkosty: oh, thanks for pointing that out!

I'm finishing a new build/deploy scripts that will pull everything from a single file (either `local.env`, `staging.env`, or `production.env`) specified as a command line arg, e.g. `./deploy.sh react --env=local`  

Once it's done we can be sure all the environment variables are consistent and you don't have to keep track of local changes to effectively 4 configuration files (`.env`, `app.yaml`, `deploys.h` - stuff like this hardcoded in it, `<project>/deploys.h` that duplicates it)

PR #94, opened 2022-09-19
Fix calendar version at build-time

realkosty:
The discrepancy between build-time version and runtime release version is causing issues not have source maps if a code was successfully built, source maps uploaded but deployment failed.

### Testing

`./deploy.sh react --env=local` -> [sample error](https://sentry.io/organizations/testorg-az/issues/3604029052/)

```
kosty@R647WWML9T-Kosty-Maleyev application-monitoring % date
Mon Sep 19 08:53:24 PDT 2022
kosty@R647WWML9T-Kosty-Maleyev application-monitoring % date -u
Mon Sep 19 15:53:27 UTC 2022
kosty@R647WWML9T-Kosty-Maleyev application-monitoring % ./release.sh
22.9.3
```


release.sh
```diff
@@ -1,9 +1,9 @@
 # This if for Release Health. It sets a release according to Calendar Versioning and gets redeployed every day
 # In sentry-demos/tracing we did not manage and autodeploy this daily, so you were stuck with the same Release for eternity until you redeployed
 # /react and /flask both have a run.sh script that utilizes this code
-day=$(date +%d)
-month=$(date +%-m)
-year=$(date +%y)
+day=$(date -u +%d)
```
thinkocapo, line 4: @realkosty you may have explained to me in Slack, but could you document in the PR what the before - vs - after is of the command here? What is the output of `$(date +%d)` versus `$(date -u +%d)`, so we can understanding?

Is it anything viewable in the UI? If you could put links to the releases. Otherwise just put the value of what `day` is.

react/src/index.js
```diff
@@ -45,7 +44,7 @@ if (window.location.hostname === "localhost") {
 
 let BACKEND_URL
 const DSN = process.env.REACT_APP_DSN
-const RELEASE = release("application.monitoring.javascript") || process.env.REACT_APP_RELEASE
```
thinkocapo, line 48: Noting I reviewed this, and I'm pretty sure this `release()` function was stale code (not needed, because process.env.REACT_APP_RELEASE is always defined, assuming you're always running ./run.sh which uses npm run build, or manually running npm run start, instead)

So good call here on simplifying.

release.sh
realkosty: thanks, I added a comment explaining it's UTC

added some testing to PR description.
I will build and run the whole thing once I have a moment too

realkosty: updated testing with sample event

PR #95, opened 2022-09-28
v7 JS SDK Upgrade, react-router v6 Upgrade, and Product/:id, Employee, Refactors

thinkocapo:
## Overview
This PR is the second attempt to upgrade from JS SDK v6 to v7. No testing of Dynamic Sampling involved anymore in this PR. When Dynamic Sampling comes back (feature flag was turned off for all orgs), v7 will be required. Session Replays also require v7.

We tried introducing JS v7 SDK last month but it caused a problem and we had to revert.

Upgraded sdk's to
**@sentry/react 7.17.3**
**@sentry/tracing 7.17.3**
**python sentry-sdk 1.10.1**

## React-Router v5->v6 - Why
React Router v6 is more widely used today. It also has a cleaner code implementation (you write `<SentryRoutes>` once and wrap routes with it, rather than wrap each individual Route with a `<SentryRoute>`). Read about the differences in config [here](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/). We'll get better support from Sentry developers if we're on the more updated version of react-router. It was also thought during debugging of Dynamic Sampling that a better react-router config (we were missing some arguments) would improve the situation, so we figured it was a good time to upgrade v5->v6 anyways.

## React-Router v5->v6 - What Changed
So we upgraded to it, and it required we turn some Smart Components into Functional Components so they would work with React Router v6. This was because the code we used for things like passing params and performing redirects,  were dependent on the version of react-router. In order for these to still work, we had to update them to how react-routerv6 requires them be done.

## Class Components to Functional Components
We updated a lot of Class Components into Functional Components, like in `Employee.js` `Product.js` `Checkout.js`

Noting that there are still some Class Components in the `/components` directory. This is okay, as now we have a mix of both and customers will be on a mix of both. Nothing inherently wrong with having both. Spending hours refactoring the rest does not have a visible effect in a demo nor provide value to the customer in a demo, so not bothering refactoring the rest at this time.

## Additional Changes Made
- useParams now used for parsing URL Params, per documentation v6 https://reactrouter.com/en/v6.3.0/getting-started/overview#configuring-routes
- no more Route Switch needed
- Redux and Context still used in some Function Components.
- useState for local state within Function Components
- useEffect for fetch'ing resources in Function Components.
- Got rid of yellow warnings from several components, which said, "React Hook useEffect has a missing dependency: 'location.state'. Either include it or remove the dependency array  react-hooks/exhaustive-deps" by moving functions into useEffect and passing values into the array it returns.
- Fixed rating calculation on product/:id page. Divide by total number of reviews and not a hard-coded amount of reviews.
- A deprecation, see next section.
- ProductCard converted from Class Component to Functional Component, this was needed as part of the deprecation.
- **Deprecated** each product json in src/react/components/products/*.json. Those were leftover from when Design Team bootstrapped this app and hard-coded the products in the frontend, as there was no backend yet (we added the backend afterwards and serve the products+reviews from there). So now, the product data is being used from the response from the backend, which has the info we need.

As a result, the Product/:is card will now have a fetch for the image, as it's pulling it front Cloud Storage now instead of local src/components/products/*.json. So the transaction shows the resource.img coming from Cloud Storage now
image

## Testing
Make sure Home, About, Products->Checkout workflows still generate the same events.

This PR is not turning on Dynamic Sampling and testing. This PR is only putting the right code into place so that DS can be tested in the future.

**Testing Dynamic Sampling in the Future**
In order for Dynamic Sampling to be configurable in the UI and not greyed out, the latest SDK versions must be present, and this means your event data must be at 5% of adoption. So we won't reach 5% unless we deploy the updated SDK's to production, because TDA is running on that 24/7. If you only run the updated SDK's locally, then the amount of times you manually load these apps will never be 5% of all the data TDA is producing, and so you'll never be able to turn on Dynamic Sampling.

## Next
Remove product images from react/src/assets, but not the About Employee images.

Can we remove `/utils/context`? Is anything still using that?

Idea: Create a run-local mode so if you don't have Backend setup, it can pull the product.js json's locally. Other developers are always requesting this. But we don't have to do it. In this case, we'd want to leave each components/product*.json in place.


realkosty:
Looks good. Assuming it works without local modifications to `requirements.txt`. Feel free to merge if tested in staging. Oh wait I'm not on approvers list.

thinkocapo:
Hold off on merging this for now. I will put the changes to db.py on a separate branch for testing Performance Issues.

Then the remainder of the code here is for converting Class Components to Functional Components and deprecating the components/products/*.json failes, which currently have many merge conflicts with Master.

Upon merging sdks-dynamic-sampling branch (has an updated react-router), these new Employee.js Product.js changes will work, as they depend on the updated react-router. @realkosty

thinkocapo:
Maybe I'll undo the db.py changes from this PR, and then rename

Performance Issues DB Query and a Product/:id Refactor
to
Product/:id Refactor

realkosty (approved):

thinkocapo:
@thinkocapo Merge and Deploy this tomorrow.

PR #98, opened 2022-10-04
Fix Performance Issues - Db query parameterization and Remove Custom Spans

thinkocapo:
## Overview
Previously we were passing strings into the Sqlalchemy ORM so if repeat queries were made, our SDK would have no way of knowing it was a N+1.

Now, we use a parameterization sqlalchemy method, and the SDK can recognize that this parameterization method is being called repeatedly, and so a N+1 query can be recognized, and Performance Issue created.

A 5th product was added to the database. I thought about doing 10 or 20 products, but that would be one awfully slow wait time for the products page to load. More thought would have to be given to how long we want the sleep statement.

There's a minimum duration of 100ms that the N+1 query must cross. Without our sleep delay in each of the 5 sql queries, we would not meet this criteria. "It needs N>=5 and a total duration of the N spans >= 100ms".

UPDATE: it's possible the custom spans would have worked, but they needed to be titled `db.<something>` and not `get_products.reviews`. So `db.get_products.reviews` would have worked.

## Small Loss of Functionality
There is no way to perform the following on the auto instrumented spans anymore, but was anyone even demo'ing it? Having span.set_tag and span.set_data working are not as significant as having Performance Issues working.

I could not get this 'tag and 'data to set propertly anymore, upon removing the custom spans and relying on the auto-instrumented ones:

```
   // Does the span.set_tag here reference the auto instrumented span from the sqlalchemy query?
    // or does it reference the custom span mentioned above it, for get_products?
    try:
        with sentry_sdk.start_span(op="get_products", description="db.connect"):
            connection = db.connect()

        n = weighter(operator.le, 12)
        products = connection.execute(
            "SELECT *, pg_sleep(%s) FROM products" % (n)
        ).fetchall()

        span = sentry_sdk.Hub.current.scope.span
        span.set_tag("totalProducts",len(products))
        span.set_data("products",products)
```
image

There's a separate engineering discussion on how to get those working for the auto instrumented span, as [Retrieve The Current Span](https://docs.sentry.io/platforms/python/performance/instrumentation/custom-instrumentation/#retrieve-the-current-span) didn't work, and it's possible the auto span is getting confused with the custom span above it for db.connect, see code.

I'm willing to live without set_tag and set_data for the time being. Performance Issues are more important ✔️.

## Testing
Not sure why 4 different performance issues (fingerprints?). It's unexplained. But creating Performance Issues works

image

and

image

Before vs After
[Before](https://sentry.io/organizations/testorg-az/discover/application-monitoring-python:ed8a2041784b4b57808da3b4a936c031/?environment=production&field=title&field=transaction.duration&field=environment&field=timestamp&name=All+Events&project=5808655&query=event.type%3Atransaction+title%3Aproducts&sort=-timestamp&statsPeriod=7d&yAxis=count%28%29) - custom spans
[After](https://sentry.io/organizations/testorg-az/discover/application-monitoring-python:6c44b35b0f854a1f86942af1f96827e4/?environment=test&field=title&field=transaction.duration&field=environment&field=timestamp&name=All+Events&project=5808655&query=event.type%3Atransaction+title%3Aproducts&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29) - no custom spans

image

## Testing on Staging
1. go here https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/products-join?se=will
2. see performance issue made https://sentry.io/organizations/testorg-az/issues/3665426046/events/a0b331a8146a4c75b1a565174473f108/?project=5808655


deploy.sh
```diff
@@ -23,4 +23,4 @@ sentry-cli deploys -o $SENTRY_ORG new -p $SENTRY_PROJECT -r $RELEASE -e $ENVIRON
 # This deploys React - The release was set in the static prod build
 gcloud app deploy --quiet
 
-cd ../flask && gcloud app deploy --quiet
+# cd ../flask && gcloud app deploy --quiet
```
realkosty, line 26: This is not a big deal at all, but I suspect this was included unintentionally so I wanted to point out you can reset your commits (`git reset` same as `git reset --mixed`) and then re-stage only the changes you want by using the interactive `git add -p` which is super neat. You can reset to previous commit with `HEAD^` or any commit and your files will remain unchanged (reset command will basically undo `git add -A` and `git commit`) so that you can selectively pick the individual chunks or changed lines to include in the commit.

so in this case I would reset all 3 commits and then interactively stage only the changes that I want like this. And then overwrite PR branch history with `--force`.
```
git reset 988da39d^
git add -p
...
git commit -m ...
git push -f
```

thinkocapo: Hmm I'm not sure I follow. I commented this out temporarily because I was deploying to Staging. I can uncomment it and commit that whenever I'm next editing code and committing, on this branch.

I guess the only one I need is the first one, I think.

I think I follow all what you're saying but I don't want to make a mistake. read through and noted all your commands offline.

I put the pg_sleep back because the engineer told me to.

realkosty:
✅LGTM

flask/db.py
```diff
@@ -46,26 +46,21 @@ def get_products():
         with sentry_sdk.start_span(op="get_products", description="db.connect"):
             connection = db.connect()
 
-        with sentry_sdk.start_span(op="get_products", description="db.query") as span:
-            n = weighter(operator.le, 12)
-            products = connection.execute(
-                "SELECT *, pg_sleep(%s) FROM products" % (n)
-            ).fetchall()
-            span.set_tag("totalProducts",len(products))
```
dgbailey, line 54: @thinkocapo just out of curiosity did logging `span = sentry_sdk.Hub.current.scope.spans` return any clues here as to what was currently on scope if anything?

I don't demo this myself so I think it is fine to move forward here.

dgbailey (approved):
@thinkocapo I looked at the grouping hashes for some of the performance issues you shared and it appears that the source span hash ( as opposed to parent or N+1 ) differs. Source span is defined in our internal docs as the the database query that happened immediately before the repeating spans. So this points towards different sleep times in the source spans as the cause. We do have some external docs around this as well https://docs.sentry.io/product/issues/performance-issues/n-one-queries/#span-evidence

realkosty (approved):
Tested in staging with standard error flow
Verified change based on latest upstream (no merge concerns)

flask/db.py
thinkocapo: I didn't try logging it but I remember setting the _tag and _data would not work anymore, once we get rid of the custom span. There becomes a confusion if it's referring to the preceding span for db.connect() or is it for the auto instrumented span for sqlalchemy SELECT * FROM PRODUCTS. @dgbailey

PR #99, opened 2022-10-04
[IGNORE] A more realistic error throwing exception for checkout

gid-sentry:
Instead of calling Sentry sdk directly which can leave the wrong impression during demos I think it would make sense to just throw an error straight out. This also demonstrates how our SDK catches errors automatically without having to manually instrument every exception to be sent up.

WIP due to testing needed to be included, not completed as of right now, though preliminary tests were good:
https://sentry.io/organizations/testorg-az/performance/trace/95053978f7154dc2b68ae29aae323b4b/?pageEnd=2022-08-31T05%3A42%3A38.043&pageStart=2022-08-30T05%3A42%3A37.903

Tracing not broken and errors showing up on both front end and back end.

To be confirmed with the few changes I made after this test.


realkosty:
Hey Gideon,

> Instead of calling Sentry sdk directly which can leave the wrong impression during demos I think it would make sense to just throw an error straight out. This also demonstrates how our SDK catches errors automatically without having to manually instrument every exception to be sent up.

I agree with that. 

1. Let's first check with @thinkocapo who knows this code in and out to make sure there are no unforeseen consequences from this change.
2. Right now you are catching the error and not re-throwing it. I don't think any errors will be sent to Sentry.

thinkocapo:
@realkosty I"ll make sure that tracing still works (looks good so far) and it still re-routes to the /error page and presents the user with "We're having difficulty", make sure nothing funky appears either in the event stream. I'll check each related event to this workflow

@gid-sentry can you put a link to a deminified error? Looks like you got something to submit but I can't see the stacktrace.

thinkocapo:
Does transaction.finish() need to be called before any redirects, or else it’s never going to get called, and the 4 events (frontend Txn, backend Txn, frontend Error, backend Error) won’t all display under 1 Trace?

thinkocapo:
Hold off on this PR, see this one instead for the fix
https://github.com/sentry-demos/application-monitoring/pull/106

react/src/components/Checkout.js
```diff
@@ -72,18 +72,30 @@ class Checkout extends Component {
...
-    if (!response.ok) {
-      this.props.history.push('/error', {"error": "errorInfo"})
-    } else {
-      this.props.history.push('/complete', {"complete": "completeOrderInfo"})
-    }
+
+    // try{
+      
+      this.setState({...this.state,loading: false});
+      transaction.finish(Date.now()+5000);
```
realkosty, line 79: Do you remember what was this about?

realkosty: `Date.now()+5000`

thinkocapo: I don't, but highly recommend removing any date/time manipulations with transactions. Risky business, leads to unwanted behaviors (broken traces, traces that don't look right).

Responding super late here so I think this is being ignored already. @realkosty

realkosty:
Closing in favor of try-throw-catch-captureException() approach:
https://github.com/sentry-demos/application-monitoring/pull/146

Had no luck with truly unhandled exception. Probably don't need it for demo purposes anyway.

PR #102, opened 2022-10-19
New build/deploy system and env-config

realkosty:
`deploy.sh` - build, upload sourcemaps, and deploy or run locally any one or multiple projects
 using `./env-config/*.env` for both build-time and runtime environment configuration. Must be
 called with `--env` parameter that specifies which env-config file to use. A special case is
 `--env=local` which will run each project's 'run.sh' script instead of deploying to Google App
 Engine (default). 
 
Example usage:
``` 
./deploy.sh react flask --env=staging
```

 Can be run from any directory. In `--env=local` mode it will start web servers for ALL projects
 specified on command-line (backends first) and then eventually, upon exit (Ctrl+C), every one 
 of them. An error anywhere in the script (e.g. failed deployment or `sentry-cli` call) should 
 make the entire script exit immediately.

 All variables in `*.env` are passed into each projects runtime environment. 
 Some variables, however, are special and are additionally used during the build:
 
   `SENTRY_ORG` is passed to `./bin/sentry-release.sh` and used to specify which Sentry
   org to create release in.

   `<PROJECT>_SENTRY_PROJECT` is passed to `./bin/sentry-release.sh` and used to specify 
   which Sentry project to create release in. 
 
   `<PROJECT>_RELEASE_PACKAGE_NAME` is used to create a new Sentry release (prepended to calendar
   version e.g. 'package-name@22.9.5' )

   `<PROJECT>_SOURCEMAPS_URL_PREFIX` 

   `<PROJECT>_SOURCEMAPS_DIR`

   `<PROJECT>_APP_ENGINE_SERVICE `is used to parametrize app.yaml.template for that project and
   points to the Google App Engine (GAE) "service" to which the project should be deployed.
   These values should be configured once in production.env and staging.env and never changed
   again. Obviously, for local.env they are meaningless and not required.
   e.g.: `REACT_APP_ENGINE_SERVICE`
 
   `REACT_APP_<PROJECT>_BACKEND `besides being passed into React application runtime these are
   substituted with the values of respective `REACT_APP_<PROJECT>_BACKEND_LOCAL` when `--env=local` 
   AND <PROJECT> is included in the list of projects to build from command-line arguments.
    
 MIGRATION NOTE: 

   The old .env and app.yaml must now be deleted from `./<project>/` directory. Instead move all the 
   variables to `./env-config/local.env` and create obtain `staging.env` and `production.env` that will be distributed 
   internally and will be added to local repository once and not require modification.


ndmanvar (approved):

ndmanvar:
amazing 😄

ndmanvar:
I'd also like for @thinkocapo to review just so we have multiple folks that have expertise the in the build/release + deploy (updated) setup.

thinkocapo (approved):

PR #106, opened 2022-11-02
feat: add api layer

scefali:
This PR adds an API layer that handles error catching that's a bit more realistic than what we have in place. Note I didn't update every usage of fetch yet. I also pinned Node and NPM versions via Volta.

### DIFF
--- a/react/package-lock.json
+++ b/react/package-lock.json
@@ -1350,6 +1350,14 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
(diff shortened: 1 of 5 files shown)


thinkocapo:
## Testing
[Error Event from 11/09/22](https://sentry.io/organizations/testorg-az/issues/3712828856/events/8464d28f6d8d44efbe1168d690182bd3/?project=5808623)
image

realkosty:
Discussed with Will, we're going hold off on this until Suspect Commits v2 is re-enabled for our org so that we know what are the implications of this to SCv2

PR #112, opened 2022-11-08
[react] Revert to busy sleep in /Products constructer in attempt to get p50 back to 8 seconds

realkosty:
Testing: none

### DIFF
--- a/react/src/components/Products.js
+++ b/react/src/components/Products.js
@@ -14,6 +14,7 @@ class Products extends Component {
 
   constructor() {
     super();
+    busy_sleep(1000);
   }
 
   // getProducts handles error responses differently, depending on the browser used
@@ -23,8 +24,6 @@ class Products extends Component {
       [ se, customerType ] = [scope._tags.se, scope._tags.customerType ]
       email = scope._user.email
     });
-
-    await sleep(1000);
 
     ['/api', '/connect', '/organization'].forEach((endpoint) => {
       fetch(this.props.backend + endpoint, {
--- a/react/src/utils/time.js
+++ b/react/src/utils/time.js
@@ -3,6 +3,16 @@ export async function sleep(milliseconds) {
   await new Promise(resolve => setTimeout(resolve, milliseconds))
 }
 
+// setTimeout is non-blocking so don't use it
+export function busy_sleep(milliseconds) {
+  var start = new Date().getTime();
+  for (var i = 0; i < 1e7; i++) {
+    if ((new Date().getTime() - start) > milliseconds){      
+      break;
+    }
+  }
+}
+
 // Consistent with "weeks" as defined in release.sh for calendar version so that
 // in Trends improved/regressed change time matches release.
 export function isOddReleaseWeek() {


thinkocapo:
@realkosty @ndmanvar is this the first time the 1 second sleep in the frontend Products component was introduced? I thought we did it a while ago. Personally I'm fine with skipping this feature as it's not realistic at all, teh developers told me you wouldn't get a random delay and wide span in the Constructor there. I remember ChrisS and I were against it. If it's causing confusion now in Trends, /products, /products-join, I'd say remove it altogether. Personally I never call it out in a demo and idk that we ever educated the team on it either.

Let's cover all this on a call. I'm not following.

realkosty:
> @realkosty are you putting back code that used to exist?
Yes, this is exactly how it was before except it was called `sleep`, I just renamed it to `busy_sleep` to reflect what it does, and `sleep` now is async. Can't have async in constructor. 

> The /products and /products-join pages and routes used to work fine - can we avoid leveraging them for Trends. This is causing a lot of confusion now. I have no idea why the P50 is changing.
> @ndmanvar too many cooks in the kitchen on the /products and /products-join workflows. We can't risk /products-join having Trends and variance now because then it could show as a longer transaction than /products which has the N+1 query and should always be slower.
> We have many separate /page(s) in the app so we can isolate different types of features and problems.

👍 I agree. I was just trying to save some work by not having to add new routes and modify TDA but I think you're right, let's not mess with it for now.

thinkocapo:
"No, we had it all along. I simply moved it out of the constructor in [Trends PR](https://github.com/sentry-demos/application-monitoring/commit/f4885719d9411ff5f5a4a045d0616b6debf73485)."

Hey @realkosty did you see the new spans when moving the sleep out of the Constructor?

If I recall correctly, it needs to be inside the constructor, so that the `react.ui.mount` span would be wider (that's the behavior @ndmanvar  wanted, for showing a true Frontend Performance problem),  and it wouldn't give us the desired span if we put it in something like `componentDidMount` instead. Looks like it got moved to the getProducts function call?

Is it non-blocking? I thought the function I was using, had some comment about being careful with how we do setTimeouts.
I wonder if this is why the Distributed Tracing is currently broken. I will investigate.

image

thinkocapo:
ok I see the `busy_sleep` makes sure it blocks program execution 👍

thinkocapo:
This is the wide ui.react.mount span I'm talking about (Frontend Slowness Problem), which I believe is dependent on the constructor() having the `busy_sleep` statement:

[/products transaction](https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:db4122cdba03468bb0c1e3f0b75470f0/?display=default&end=2022-11-03T10%3A49%3A39.000&field=title&field=backendType&field=transaction.duration&homepage=true&name=All+Events&project=5808623&query=se%3Atda+event.type%3Atransaction+title%3A%2Fproducts+backendType%3Aflask&sort=-title&start=2022-11-03T02%3A17%3A57.000&yAxis=count%28%29) from Nov3rd
image

thinkocapo:
the video also demonstrates that the `busy_sleep` needs to be in the constructor, for the react.ui.mount to be a wide span.

I'll open up a PR for the fix. My PR does not address the "p50 back to 8 seconds" but restoring Distributed Tracing is more urgent, for the time being.

thinkocapo:
Actually do those transactions with the messed up p50, maybe those are also the ones where Distributed Tracing is broken?? Maybe the next PR i'm making, will fix it.

thinkocapo:
Ok need to merge this PR asap, for Distributed Tracing to work again, and it might solve the P50 duration problem too. If not, we can address P50 duration problem afterwards.
https://github.com/sentry-demos/application-monitoring/pull/113

PR #119, opened 2022-11-10
Fixes Express DB Config and Broken Boolean Evaluation of Headers

thinkocapo:
PUBLIC_CLOUD_IP is not a var in config, but HOST already is (it's the IP address that PUBLIC_CLOUD_IP was trying to refer to) and HOST is already available in config.

Angelo and I found that if you pass `undefined` as the value (sometimes the calling Frontend doesn't pass it, or you're pasting the URL into the Browser URL bar and not setting headers), then this still evaluates to true, and it sets `customerType:undefined` and so express's /products's http call to Ruby fails on 
```
Server process(es) are running. Press Ctrl+C to terminate...
> /products headers { se: undefined, customerType: undefined, email: undefined }
TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "se"
    at ClientRequest.setHeader (_http_outgoing.js:536:3)
    at new ClientRequest (_http_client.js:249:14)
    at Object.request (https.js:310:10)
```
^ notice the headers are still undefined, that is a result of the following code:

```
  if(!['undefined'].includes(customerType)) {
    Sentry.setTag("customerType", customerType);
    headers["customerType"]=customerType
  }
```
^ This is poor readability, the boolean evaluation, "if not undefined includes customerType", is confusing.

So far it's fixed by, "if customerType is not undefined"
```
  if(customerType !== undefined) {
    Sentry.setTag("customerType", customerType);
    headers["customerType"]=customerType
  }
```
^ no longer sets customerType:undefined. Rather, it skips setting the header altogether, if the value coming in is undefined.

## Testing
[trace link](https://sentry.io/organizations/testorg-az/performance/trace/2d03376df42b43a7a55b5890ce21989a/?pageEnd=2022-11-11T02%3A57%3A58.862&pageStart=2022-11-10T02%3A57%3A48.039) (includes Frontend React app transactions, headers)

Headers work when present as well as not provided.

Backend products txn in Express looks fine
Haven't tested on Staging, but neither has the new deploy.sh been tested for Staging.
No way this will reach production if it doesn't work on Staging first.

### DIFF
--- a/express/db.js
+++ b/express/db.js
@@ -142,6 +142,9 @@ function openDBConnection() {
 
   let host
   if (process.env.EXPRESS_ENV === 'test') {
+    // The cloud sql instance connection
+    // name doesn't work locally, but the
+    // public IP of the instance does.
     host = process.env.HOST
   } else {
     host = '/cloudsql/' + process.env.CLOUD_SQL_CONNECTION_NAME
--- a/express/db.js
+++ b/express/db.js
@@ -1,5 +1,6 @@
 require('dotenv').config();
 const Sentry = require('@sentry/node');
+// const { Console } = require('@sentry/node/dist/integrations');
 
 // Knex is the database query builder used in the GCP docs, which
 // is why we are using it here. See docs:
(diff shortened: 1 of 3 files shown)


Angelodaniel (approved):

realkosty (approved):
Makes sense
Ideally we'd rename HOST->CLOUD_SQL_PUBLIC_IP everywhere because HOST can conflict with React HOST, but that would break things and require everyone update all of their *.env files

PR #124, opened 2022-11-17
New Page Route for Trends

thinkocapo:
## Overview
This is for some delays to be scheduled in the Organization.js page route, for powering Trends, either an Improvement or Regression.

## Design
The /organization page doesn't call any backends, but it doesn't need to, that I know of. Because the delay is set in the frontend code and not the backend.

Note there's already a backend transaction called `organization` that's called by the backend /products as well as the frontend /about page. So could rename ORganization.js as Settings.js if we want to, something more neutral. Not sure if anyone will expect Organization.js to be calling `organization` the api endpoint in all our backends. We could also purposely make it do that.

## Testing
the new /organization pageload transactions
https://sentry.io/organizations/testorg-az/performance/summary/?project=5808623&query=transaction.op%3Apageload&referrer=performance-transaction-summary&statsPeriod=24h&transaction=%2Forganization&unselectedSeries=p100%28%29

the new /organization pageload transaction, on Staging
https://sentry.io/organizations/testorg-az/discover/application-monitoring-javascript:e350c1d440f54bd7ad00edcfd80f3bb6/?field=event.type&field=title&field=se&field=timestamp&homepage=true&name=All+Events&project=5808623&query=url%3A%22https%3A%2F%2Fstaging-application-monitoring-react%2A%22&sort=-event.type&statsPeriod=1h&yAxis=count%28%29

TDA running against the Staging app /organization
https://sentry.io/organizations/testorg-az/discover/homepage/?end=2022-11-21T15%3A58%3A00&field=event.type&field=title&field=se&field=timestamp&name=All+Events&project=5808623&query=url%3Ahttps%3A%2F%2Fstaging-application-monitoring-react%2A+title%3A%2Forganization&sort=-event.type&start=2022-11-21T15%3A55%3A12&yAxis=count%28%29


realkosty (approved):
Thanks for doing this!

PR #152, opened 2023-01-11
Add Ruby on rails service/skeleton

ndmanvar:


realkosty (approved):
🚀

env-config/example.env
```diff
@@ -13,6 +13,7 @@ REACT_APP_SPRINGBOOT_BACKEND=
 REACT_APP_ASPNETCORE_BACKEND=
 REACT_APP_LARAVEL_BACKEND=
 REACT_APP_RUBY_BACKEND=
+REACT_APP_RAILS_BACKEND=
```
realkosty, line 16: This needs to be consistent with `ruby-on-rails` -> `RUBYONRAILS` for deploy.sh local port rewiring to work
https://github.com/sentry-demos/application-monitoring/blob/262aa3ec3b7f06605ce202acd3e611f7dd13ffa2/deploy.sh#L144

env-config/example.env
```diff
@@ -72,3 +73,11 @@ RUBYONRAILS_LOCAL_PORT=3000 # only needed in local.env
...
 RUBYONRAILS_SECRET_KEY_BASE=
 
+RUBYONRAILS_APP_DSN=
+RUBYONRAILS_RELEASE_PACKAGE_NAME=
+RUBYONRAILS_APP_ENGINE_SERVICE=application-monitoring-rubyonrails
+RUBYONRAILS_SENTRY_PROJECT=
+RUBYONRAILS_LOCAL_PORT=3000 # only needed in local.env
+SECRET_KEY_BASE=
+RAILS_ENV=
+PIDFILE=
```
realkosty, line 83: 1. this seems like it already existed in the file right above (lines 68-74)
2. [nit] I'm personally inclined to hardcode PIDFILE in app.yaml.template as implementation detail and not something that we will ever change or affect data in Sentry

Also (mostly curiosity) what's SECRET_KEY_BASE? I can't seem to find where it's used

realkosty (changes requested):

env-config/example.env
realkosty: and if you don't mind let's keep all vars prefixed like you had originally for consistency

ndmanvar: got it 👍 ill make these changes today. thanks for the review @realkosty

ndmanvar: @realkosty - re `SECRET_KEY_BASE` -> https://cloud.google.com/ruby/rails/appengine#configure_the_rails_secret_key

realkosty: thanks!!

env-config/example.env
realkosty: @ndmanvar this should be `REACT_APP_RUBYONRAILS_BACKEND=` because the script automatically derives it from `ruby-on-rails` by removing dashes and capitalizing, kind of like with the variables prefixed with `RUBYONRAILS_`

ndmanvar: done. made the changes in `-config` 
https://github.com/sentry-demos/application-monitoring-config/pull/1/commits/4b6dfc4977b71b56f316a290efb878b52630be21

PR #153, opened 2023-01-12
Update README with setup instructions

katiejenkinswebdev:


README.md
```diff
@@ -16,9 +16,16 @@ Also called the Empower Plant UI/UX. This project was bootstrapped with [Create
 | sentry-logback | 5.5.1 |
 
 ## Setup
-1. Permit your IP address in CloudSQL.
+1. Permit your IP address in CloudSQL by posting a message in the `#se-team` Slack channel.
```
katiejenkinswebdev, line 19: Talked to Dustin & Simon about pointing out a Slack channel here.

README.md
```diff
@@ -121,9 +128,9 @@ git merge master
 
 # update sentry_sdk's and other modules
 cd react && rm -rf node_modules && npm install
-cd flask && source env/bin/activate && pip install -r requirements.txt
+cd flask && source env/bin/activate && pip3 install -r requirements.txt
```
katiejenkinswebdev, line 131: Updated to newest command

README.md
realkosty: Let's make it:
1. Permit your IP address in CloudSQL
    - Go to https://console.cloud.google.com/sql/instances
    - You will see 1 instance, under Actions column click "..." -> Edit
    - Expand "Connections", under Authorized Networks click "ADD NETWORK"
    - Google "my IP address", add it.

README.md
```diff
@@ -16,9 +16,16 @@ Also called the Empower Plant UI/UX. This project was bootstrapped with [Create
...
-4. The `REACT_APP_FLASK_BACKEND` in `env-config/*.env` points to the backend instance deployed to AppEngine, the same one used by the cloud-hosted React web app. Flask is the default backend. If you expect to run other backend types, add values for those in `env-config/*.env` as well (i.e. `REACT_APP_EXPRESS_BACKEND`).
+3. The `REACT_APP_FLASK_BACKEND` in `env-config/local.env` points to the backend instance deployed to AppEngine, the same one used by the cloud-hosted React web app. Flask is the default backend. If you expect to run other backend types, add values for those in `env-config` in your `local.env` file as well (i.e. `REACT_APP_EXPRESS_BACKEND`).
+4. Confirm [Homebrew](https://brew.sh/) is installed with `brew -v`. If not, install using `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`.
+5. Confirm PostgreSQL is installed with `Postgres -V`. If not, install using `brew install postgresql`.
```
realkosty, line 23: nice!

README.md
```diff
@@ -16,9 +16,16 @@ Also called the Empower Plant UI/UX. This project was bootstrapped with [Create
...
+6. As the application is compatible with specific versions of `node` and `npm`, install the following to be able to set the specific versions required:
+    1. Install compatible `npm` version with `npm install -g npm@XX.XX.XX`. NOTE: may need to use `sudo` with command.
+    2. Install `n` to update `node` version with `npm install -g n`.
+    3. Set the specific `node` version with `n XX.XX.XX`. NOTE: may need to use `sudo` with command.
```
realkosty, line 27: cool, I didn't know about `n`
I used `nvm` to do that. Is `n` preferable?

README.md
```diff
@@ -31,14 +38,14 @@ Also called the Empower Plant UI/UX. This project was bootstrapped with [Create
...
-NOTE: `build.sh` and `run.sh` files in each project are not meant to be run direclty, use top-level `deploy.sh` instead because it sets all required environment variables.
+NOTE: `build.sh` and `run.sh` files in each project are not meant to be run directly, use top-level `deploy.sh` instead because it sets all required environment variables.
 
 If you run locally and only deploy `react` it will automatically point to `staging` backends, however if you include a backend
 projects in the command `react` will magically point to it instead of staging (still requires `&be=<backend>` url param).
 
 `deploy.sh` takes another argument `--env=<env>`, which can be either `local`, `staging` or `production`. Each value corresponds to a file in `env-config` directory. `local` is a special value, most significantly it will run all webservers locally instead of deploying to Google App Engine.
 
-`deploy.sh` does everything including validating that all required values are set in the `env-config/*.env`, that each project's
+`deploy.sh` does everything including validating that all required values are set in the `env-config/local.env`, that each project's
```
realkosty, line 48: Lets keep it the way it is because in addition to `env-config/local.env` this applies to all of 
```
env-config/staging.env
env-config/production.env
env-config/my-test-env.env
```

README.md
realkosty: thanks

README.md
```diff
@@ -121,9 +128,9 @@ git merge master
 
 # update sentry_sdk's and other modules
 cd react && rm -rf node_modules && npm install
-cd flask && source env/bin/activate && pip install -r requirements.txt
+cd flask && source env/bin/activate && pip3 install -r requirements.txt
 
-# Check that your react/.env, flask/.env and deploy.sh still have correct values
+# Check that your react/local.env, flask/local.env and deploy.sh still have correct values
```
realkosty, line 133: Let's remove this line entirely. Back in the day we would change values in `react/.env`, now this is no longer necessary

README.md
katiejenkinswebdev: ooo I like this even better! Updating now...

README.md
katiejenkinswebdev: That makes sense - updating now!

README.md
katiejenkinswebdev: I've used both! Let me see which one is more preferred. Looking into now...

katiejenkinswebdev: Here are the main differences between `n` and `nvm`. I'm open to changing this to `nvm` since it is more widely known. Let me know what you think is best!
```
The main advantage of n's approach to managing Node.js releases compared to NVM is that global npm packages are not affected when you switch between different versions of Node. On the other hand, NVM lets you use different Node.js versions in different terminals through its nvm use command, while n does not. Once you switch to a different Node.js version, the change is reflected system wide. You can get around this by using the n use subcommand to invoke a specific Node.js binary without switching to that version. This can come in handy when performing one-off script executions.
```
https://www.honeybadger.io/blog/node-environment-managers/

README.md
katiejenkinswebdev: Done!

README.md
realkosty: Thanks for the summary, that's good to know. Let's leave `n` since you already tested those commands. I'll add a "alternatively you can use nvm" some other time.

realkosty (approved):
Thank you for improving this!

PR #177, opened 2023-02-22
Adds N+1 Functionality 

bfpepper87:
This adds the ability for the React app to showcase the N+1 API Calls functionality. The 20 calls that are made need to be that high to trigger the event as confirmed with George in engineering.  Nothing is needed from the user. Go to /nplusone and the rest is taken care of. The back end in Flask is the one handling the backend call. 

Test: 

Event in Sentry. 
https://testorg-az.sentry.io/issues/3952653497/?query=is%3Aunresolved+issue.type%3Aperformance_n_plus_one_api_calls&referrer=issue-stream&statsPeriod=14d


flask/main.py
```diff
@@ -150,6 +150,13 @@ def organization():
 def connect():
     return "flask /connect"
 
+@app.route('/product/0/info', methods=['GET'])
+def product_info():
+    import time
```
thinkocapo, line 155: could you move this to the top of the file where the other `import` statements are?

flask/main.py
```diff
@@ -150,6 +150,13 @@ def organization():
...
+    time.sleep(.55)
+    response = make_response(generateMetrics(), 200)
+    response.mimetype = "text/plain"
```
thinkocapo, line 158: If `make_response(...)` on the previous line is what sends the response, then is `response.memtype` doing anything after the fact? Should it be before `response = make_response(...)`? Or if you could message me the code example you found this on, I can take a look.

react/src/components/nplusone.js
```diff
@@ -0,0 +1,53 @@
...
+class Nplusone extends Component {
+  constructor(props) {
+    super(props);
+    const ids = new Array(20).fill(0).map((item, index) => index);
+    this.state = {
+      ids,
+    };
+  }
+  async componentDidMount() {
+    console.log("Mounting NPlusOne with props", this.props);
```
thinkocapo, line 31: Hey could you remove all the other console.log's you committed? I'm okay with this one in componentDidMount, just so we have a log that the component is properly mounting.

react/src/components/nplusone.js
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <div>
+        <h1>API N+1 Performance Issue</h1>
+        {this.state.ids.map((index) => (
+          <NPlusOneItem key={index} index={index} backend={this.props.backend} />
+        ))}
+        <h3>
+          look at the query{" "}
+          <a
+            href="https://testorg-az.sentry.io/issues/?project=5808623&query=is%3Aunresolved+issue.type%3Aperformance_n_plus_one_api_calls&referrer=issue-list&statsPeriod=14d"
```
thinkocapo, line 43: Nice Ben, a link back into Sentry,io. Reminds me of how @realkosty put links back to Sentry,io in the TDA testing logs as they complete.

Good idea maybe we'll re-use more in the future! Saves time from having to go find and setup the query for the data.

react/src/components/nplusone.js
```diff
@@ -0,0 +1,53 @@
...
+import { determineBackendUrl } from "../utils/backendrouter";
+class NPlusOneItem extends Component {
+  async componentDidMount() {
+    console.log("Mounting NPlusOneItem");
+    const response = await fetch(this.props.backend + "/product/0/info?id=" + this.props.index);
+    console.log({ response });
+    const json = await response.json();
+    console.log({ json });
+    this.setState({
+      data: json,
```
thinkocapo, line 15: Does this `data` get used anywhere? I think not, and that's not necsarilly a problem. But I want to be aware what happens if a bunch of 500 responses get sent back (let's say the backend is down or just screwing up) - then what do we expect to happen with the NPlusOne component? Will it still render?

If it still renders, then that can be fine too, but dunno if it will stop the Perf Issue from occurring or any other erroneous outcomes we'll stumble upon later.

react/src/components/nplusone.js
```diff
@@ -0,0 +1,53 @@
+import { Component } from "react";
+import Context from "../utils/context";
+import { Link, useFetcher } from "react-router-dom";
+import * as Sentry from "@sentry/react";
+import { connect } from "react-redux";
+import { determineBackendUrl } from "../utils/backendrouter";
+class NPlusOneItem extends Component {
+  async componentDidMount() {
+    console.log("Mounting NPlusOneItem");
+    const response = await fetch(this.props.backend + "/product/0/info?id=" + this.props.index);
```
thinkocapo, line 10: Hey, in the event this errors (say a 500 response) - what happens to the NPlusOne component? Does it still render? Let's get some basic try/catch Sentry.captureException() style error handling in here.

Any chance you could test sending back a 500 response from the Flask end so we can see how it responds?

It's a lot of work now, but it will save us tons of headache debugging time if Flask is ever down or unexpected outcomes happen in the future.

flask/main.py
bfpepper87: So I'm pretty sure I remember seeing a 200 response from my initial tests. However, today I'm getting back 500s regardless of what Flask is set to (tried with both 200 and 500). Regardless of what the response should be, the N+1 API call still works so, from my reading of the docs and the behavior I'm seeing, the response doesn’t matter only that the call fulfills the detection criteria.

flask/main.py
bfpepper87: Moved

react/src/components/nplusone.js
```diff
@@ -0,0 +1,53 @@
...
+    console.log({ response });
+    const json = await response.json();
```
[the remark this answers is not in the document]
bfpepper87: Its not really needed so im just removing the json part.

react/src/components/nplusone.js
bfpepper87: Ya. We get back 500s since there isnt anything actually being sent back. but the page loads and the perf event triggers.

react/src/components/nplusone.js
bfpepper87: Yes. Currently getting 500s from the backend and the front end renders properly. I think this is ok.

thinkocapo (approved):

deploy.sh
```diff
@@ -165,7 +165,7 @@ for proj in $projects; do # bash only
   ./build.sh
 
   if [[ "$fe_projects" = *"$proj "* ]]; then # project is frontend
-    sentry-release.sh $env $RELEASE
+    bash -x sentry-release.sh $env $RELEASE
```
realkosty, line 168: Hey Ben, this should not have been included in commit
same for removing env-config/example.env, which also removed directory and as a result actually [broke auto-deployment](https://github.com/sentry-demos/application-monitoring/commit/2372580b68de72196c06c3ecf05a3111a1ffb8eb)
No big deal, the way to avoid it next time is not to use `git add -A`, very common but a bad practice
I'm gonna revert the 2 unwanted changes tomorrow.
@thinkocapo fyi

realkosty: no worries, fixed it
https://github.com/sentry-demos/application-monitoring/commit/767f43fff47b806d282cf9d8b396e1b78371bff0

react/src/components/nplusone.js
thinkocapo: @bfpepper87 can you try to add some better erroring handling for all these failed responses that are appearing? If we could group them together that'd be sweet. No stacktraces on these so hard to make sense of what's happening. Not sure of it's the outbound http request or the http Response yet.
https://testorg-az.sentry.io/discover/homepage/?field=issue&field=title&field=count%28%29&name=All+Events&project=5808623&query=event.type%3Aerror+http.url%3Ahttps%3A%2F%2Fapplication-monitoring-react-dot-sales-engineering-sf.appspot.com%2Fnplusone&sort=-count&statsPeriod=7d&yAxis=count%28%29

react/src/components/nplusone.js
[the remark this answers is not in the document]
thinkocapo: @bfpepper87 what is the expected response? this is erroring, not sure what to expect.

flask/main.py
```diff
@@ -150,6 +151,10 @@ def organization():
 def connect():
     return "flask /connect"
 
+@app.route('/product/0/info', methods=['GET'])
+def product_info():
```
thinkocapo, line 155: @bfpepper87 does this provide a response? I don't see one, but maybe I'm missing something. This is erroring at scale https://testorg-az.sentry.io/issues/3966708105/events/5aee332b1cd441b0aa76effc1807f9d7/?project=5808655 can you fix?

PR #183, opened 2023-03-07
Add slow profile python

sergiosentry:
## Type of Change

[] Bugfix
[] New Feature
[x] Enhancement
[] Refactoring

## Description

Add slow profile method for python service.
Whenever we get a list of plants from the database, we are going to check if those plants have pests in the description. If they do, we will remove them from the list. 
This is a dumb method whose purpose is only to introduce slowness into this method so we can build a story around how the `get_products` method is really slow and profiling gives us the context we need to figure out where the problem is coming from. This method calls the `get_iterator` function which is basically a Fibonacci code in disguise.

It adds about 2 seconds to the load time. **This is visible from the front-end applications when loading the products page**

## Testing

[Normal profile (Without slow method)](https://testorg-az.sentry.io/profiling/profile/flask-backend/24992bb643164deebd28212cd681469c/flamechart/?colorCoding=by+symbol+name&query=&sorting=call+order&tid=10754224128&view=top+down)

[Slow profile (With slow method)](https://testorg-az.sentry.io/profiling/profile/flask-backend/6a3f7ff7a4df464b97f1f27cd25ad7c7/flamechart/?colorCoding=by+symbol+name&query=&sorting=call+order&tid=6238842880&view=top+down)


thinkocapo:
Hey the profiles in Testing look good. get_iterator is shown in the transaction too.

Top - normal txn (without slow method)
Bottom - slow txn (with slow method)

image

PR #185, opened 2023-03-13
[tda]: Refactor tests (homepage only run crash, products->checkout, + /nplusone

ndmanvar:


ndmanvar:
https://testorg-az.sentry.io/issues/3952653497/?end=2023-03-13T23%3A43%3A15&query=se%3Aneilmanvar+%21project%3Ajob-monitor-application-monitoring&referrer=issue-stream&start=2023-03-13T23%3A42%3A13
image

ndmanvar:
test_checkout session-replay: https://testorg-az.sentry.io/replays/application-monitoring-javascript:31ef969da2004ac797139a538939a424/?referrer=%2Fissues%2F%3AgroupId%2F&t=11&t_main=console
- starts on `/`
- clicks on products btn to get to `/products`

ndmanvar:
@realkosty - ready for your review! (:

tests/desktop_web/test_products_join.py
```diff
@@ -2,8 +2,8 @@
 import sentry_sdk
 from urllib.parse import urlencode
 
-def test_checkout_products_join(desktop_web_driver, endpoints, random, batch_size, backend, sleep_length):
-    sentry_sdk.set_tag("pytestName", "test_checkout_products_join")
+def test_products_join(desktop_web_driver, endpoints, random, batch_size, backend, sleep_length):
+    sentry_sdk.set_tag("pytestName", "test_products_join")
```
realkosty, line 6: sweet!! no more misnamed tests

I think in the future it would also be nice to have a comment at the top of each test
describing what purpose it serves (e.g. this one would be "demo transaction spans to show AFTER fixing N+1 Q issue")

tests/desktop_web/test_performance_issues.py
```diff
@@ -0,0 +1,19 @@
+import time
+import pytest
+import sentry_sdk
+
+def test_nplusone(desktop_web_driver, endpoints, random, batch_size, sleep_length):
```
realkosty, line 5: So I know you're just following existing convention but wondering if we should rename both to `nplusone_api_call` to disambiguate.

tests/desktop_web/test_performance_issues.py
```diff
@@ -0,0 +1,19 @@
...
+    sentry_sdk.set_tag("pytestName", "test_nplusone")
+
+    for endpoint in endpoints['react_endpoints']:
+        endpoint_organization = endpoint + "/nplusone"
+        sentry_sdk.set_tag("endpoint", endpoint_organization)
+
+        url = endpoint_organization
+
+        for i in range(batch_size):
```
realkosty, line 14: [POSSIBLY REQUESTING CHANGE]

Not part of this PR, but

```
        for i in range(batch_size):
```
will generate batch_size * 20 hits to `flask`, so average of 200 requests. `/nplusone` right now only sleeps but still it's a lot of threads/processes. This could become a problem and require us to scale up flask GAE service.

on the other hand we want performance issues to show up high up-ish in the issue stream...

I'd suggest running this with BATCH_SIZE=40 (2x current max batch size) and seeing if anything abnormal happens and if it does change this to:
```
for i in range(max(1, batch_size//2)):
```

realkosty (changes requested):

realkosty:
> test_checkout session-replay: https://testorg-az.sentry.io/replays/application-monitoring-javascript:31ef969da2004ac797139a538939a424/?referrer=%2Fissues%2F%3AgroupId%2F&t=11&t_main=console
> 
> * starts on `/`
> * clicks on products btn to get to `/products`

That user must be on crack or they really know what they want. From 0 to spending $350 in 11 seconds ⚡️

tests/desktop_web/test_performance_issues.py
ndmanvar: yep, renaming 👍

tests/desktop_web/test_performance_issues.py
ndmanvar: understood. Making GH issue with this in it

ndmanvar: https://github.com/sentry-demos/application-monitoring/issues/186

tests/desktop_web/test_products_join.py
ndmanvar: ah good point. I'll get that done (at some point)

ndmanvar:
@realkosty - addressed all comments.
Can you re-review when you get a chance?

realkosty (approved):

PR #190, opened 2023-03-24
Katie laravel demo

katiejenkinswebdev:


realkosty:
Nice. Does this actually end up querying the `inventory` table in DB?

ndmanvar (approved):

PR #193, opened 2023-03-28
Ruby on rails gw

gywright:
added rails backend.
tested live with Neil.


realkosty (changes requested):
Amazing. Just 1 minor comment

ndmanvar:
commited the rails tda change directly conftest

PR #194, opened 2023-03-29
React uncompressed asset + fix /product/0/info + add tracePropagationTargets

ndmanvar:


flask/main.py
```diff
@@ -178,18 +178,37 @@ def connect():
 @app.route('/product/0/info', methods=['GET'])
 def product_info():
     time.sleep(.55)
+    return "/product/0/info"
```
ndmanvar, line 181: will fix this error that we are seeing a bunch which messes with our crash-free session rate
https://testorg-az.sentry.io/issues/3966708105/?project=5808655&referrer=release-issue-stream
image

image

PR #207, opened 2023-04-27
add sleep to android checkout test

sdzhong:


ndmanvar:
let's eventually replace this by waiting until the pop up dialog is rendered
image

image

ndmanvar (approved):

ndmanvar:
i have a feeling this happens cause the test execute "too fast" locally, as opposed to when running on VM in a bunch of threads due to parallel testing

sdzhong:
Other changes are being discussed. Instead of a hardcoded number for sleep (`time.sleep(2)`), we can use other options, like waiting on the element via XPATH:
WebDriverWait(android_emu_driver, 5).until(EC.presence_of_element_located((AppiumBy.XPATH, "/hierarchy/android.widget.FrameLayout/android.widget.FrameLayout/android.widget.FrameLayout/")));

Although instead of XPATH, we would prefer to get the element via `AppiumBy.ID`. This is an active work-in-progress such that this PR may be irrelevant.

ndmanvar:
image

PR #212, opened 2023-05-12
Overhaul ASP.NET Core backend

mattjohnsonpint:
- Updates to ASP.NET Core 7
- Enables Performance (tracing), Release Health (session tracking), and source context (symbols & sources upload)
- Switched to using a custom dockerfile for GAE deployment.
- Updates the build/deploy scripts where needed.
- General cleanup and mininimalizaiton throughout.
- Uses google cloud secret manager to store the authentication token used for Sentry CLI.

Already deployed to staging.  Production should be updated after merging this.


thinkocapo (approved):
@mattjohnsonpint could you describe some of the benefits of going with the custom Dockerfile? So we're aware and don't forget this is one of our demo's using a custom dockerfile. I don't think our other apps in GAE are using such. The app.yaml is still in use yea? I'm taking a look.

aspnetcore/Controllers/CheckoutController.cs
```diff
@@ -1,48 +1,37 @@
...
-    [ApiController]
-    [Route("[controller]")]
-    public class CheckoutController : ControllerBase
+    [HttpPost]
+    public void Checkout()
     {
-        private readonly ILogger<CheckoutController> _logger;
-        private hardwarestoreContext _context = null;
-        private readonly IConfiguration Configuration;
+        var se = (string?) Request.Headers["se"];
```
bruno-garcia, line 10: Would we want this set in all events? e.g: adding this on a middleware

aspnetcore/Program.cs
```diff
@@ -1,28 +1,56 @@
...
+builder.WebHost.UseSentry(options =>
 {
-    public class Program
+    // Use the environment variables set by the startup scripts.
+    options.Dsn = Environment.GetEnvironmentVariable("ASPNETCORE_APP_DSN");
+    options.Release = Environment.GetEnvironmentVariable("RELEASE");
+    
+    // Enable some features.
+    options.TracesSampleRate = 1.0;
+    options.AutoSessionTracking = true;
```
bruno-garcia, line 37: This will only capture a session when the process starts and end it when it shuts down (client mode). Or was [server mode](https://develop.sentry.dev/sdk/sessions/#server-model) implemented?

I'm aware of some attempts but they never landed:
* https://github.com/getsentry/sentry-dotnet/pull/1144
* https://github.com/getsentry/sentry-dotnet/pull/1162

Should we go without it?

```suggestion
```

bruno-garcia (commented):
Very nice!

aspnetcore/Program.cs
mattjohnsonpint: @bruno-garcia - I brought up server-side session tracking at Client Infra TSC awhile back, and the consensus was to not worry it as a per-request feature, as that's something that Transactions are already covering.  Still, what you are calling "client mode" still actually has benefit for server-based apps, as it will show a release unhealthy if the server process crashes.   I wish we had clearer terminology than "sessions" here, because you're right - it's not the user's session, but rather the lifetime of the application.

aspnetcore/Controllers/CheckoutController.cs
mattjohnsonpint: Not sure.  I didn't investigate the end-to-end usage of the application.  The endpoints all do what they previously did.  There's still room for improvement such as this, or making an outbound http request, or connecting some of the other endpoints to return results from the database.

mattjohnsonpint:
Some events in staging, using the react front-end locally, pointed at the staging aspnetcore backend.
- [`POST /checkout` Error](https://testorg-az.sentry.io/issues/4176714561/events/9af53430e30f40f28b2345e741505ded/)
- [`POST /checkout`Transaction](https://testorg-az.sentry.io/performance/staging-application-monitoring-aspnetcore:3935b7f692184e0ea6518ae6b14b26c0/?transaction=POST+Checkout)
- [`GET /products` Transaction](https://testorg-az.sentry.io/performance/staging-application-monitoring-aspnetcore:33ec6c065a9b456ca5bb149fd36423fe/?transaction=GET+Products)

mattjohnsonpint:
Switching to draft for a moment, as I want to see if I can store the new auth token as a cloud secret instead of requiring a change to the env files.

mattjohnsonpint:
> Switching to draft for a moment, as I want to see if I can store the new auth token as a cloud secret instead of requiring a change to the env files.

Done.  The secret is retrieved from Google Cloud Secrets Manager during deployment, as long as `gcloud` utility is installed and authenticated - which it needs to be for deploying to staging or production anyway.

aspnetcore/Dockerfile.template
```diff
@@ -0,0 +1,23 @@
+﻿FROM mcr.microsoft.com/dotnet/aspnet:7.0 AS base
+WORKDIR /app
+ENV ASPNETCORE_URLS=http://+:8080
+EXPOSE 8080
+
+FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+ENV SENTRY_AUTH_TOKEN=${SENTRY_AUTH_TOKEN}
+ENV SENTRY_ORG=${SENTRY_ORG}
+ENV ASPNETCORE_SENTRY_PROJECT=${ASPNETCORE_SENTRY_PROJECT}
```
mattjohnsonpint, line 9: FYI - I hate that we have to do this - especially for the auth token.  Ideally we'd not have to templatize the dockerfile at all, and would just use `ARG` to pass through to `docker build` ([see this blog for details](https://vsupalov.com/docker-build-pass-environment-variables/)).  Unfortunately, GAE doesn't let us control the `docker build` command it uses.  There's no way to pass the arguments in.  ([see stackoverflow question and answer](https://stackoverflow.com/q/68152233/634824)).

mattjohnsonpint: Note, `env_variables` set in the `app.yaml` are passed into the docker container only at run time - in the `final` stage of this file.  They're not available during the `build` stage, which is where we need them.  (This is for getting the `dotnet build` to invoke Sentry CLI to upload symbols and sources.)

PR #231, opened 2023-05-25
Move slowness to interceptor

mattjohnsonpint:
Revise #230  - Do the slowness in a DB interceptor so it looks like the query was slow, rather than a delay before the query.


thinkocapo (approved):

realkosty:
This is much cleaner than sleepy views. 
Maybe someday we can figure out how to do something like this in all the other frameworks

PR #235, opened 2023-07-16
[release] Deploy React when time for new calendar release even if no changes

realkosty:
…


realkosty:
Merging without review to quickly verify this works since GH Actions code is a pain to test and I'm going on vacation in 24 hours.

PR #236, opened 2023-07-18
feat(profiling): add browser profiling

JonasBa:
Upgrades the SDK version and removes @sentry/tracing which is deprecated and sets up browser profiling

### DIFF
--- a/react/package-lock.json
+++ b/react/package-lock.json
@@ -8,7 +8,7 @@
       "name": "application-monitoring",
       "version": "0.1.0",
       "dependencies": {
-        "@sentry/react": "^7.59.2",
+        "@sentry/react": "^7.59.3",
         "@testing-library/jest-dom": "^5.11.4",
         "@testing-library/react": "^11.1.0",
         "@testing-library/user-event": "^12.1.10",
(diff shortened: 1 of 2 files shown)


react/src/index.js
```diff
@@ -14,6 +13,7 @@ import { createStore, applyMiddleware, compose } from 'redux'
 import logger from 'redux-logger'
 import rootReducer from './reducers'
 
+
```
ndmanvar, line 16: unnecessary new line (:

react/src/index.js
```diff
@@ -82,6 +84,12 @@ Sentry.init({
           // name: window.location.pathname.replace(/\/employee.*/,'/employee/:id')
         };
       },
+      _experiments: {
+        // This enables tracing on user interactions like clicks
+        enableInteractions: true,
```
ndmanvar, line 89: Hmmm, I'll have to check that this doesn't break our custom-instrumented checkout transaction that has an error attached to it.

ndmanvar (approved):

ndmanvar (changes requested):
Profiles are not source mapped.
Let's get that source mapped before this PR is approved / ready to be merged.

ndmanvar (approved):

PR #237, opened 2023-07-20
replay: capture network bodies

bruno-garcia:


realkosty (approved):
Thanks!
Verified request body present on /checkout in Replay view

[flagship transaction](https://testorg-az.sentry.io/performance/staging-frontend-javascript:cbfd08f71d6c4e52a0d3edbf0b47fcab/) has existing regression - no (+) next to /products request span even though BE request linked through distr. tracing. Verified it happened before this change by looking at current production events.

PR #245, opened 2023-08-04
feat(react): render problem

JonasBa:
Array length of 250/4 is a float hence the invalid array length. Lower it down to 200

PR #249, opened 2023-08-07
Revert "feat(react): render problem (#245)"

realkosty:
This reverts commit 6fc72e26a5820ae3492ac3611da73a1ffd985bb5.
Something changed about the cart or something else in the /checkout payload which made checkout succeed in some of the backends. Before we make changes to all backends so that they fail regardless of request body we need to temporarily roll this back so that demo flow is not broken for SEs.

# Testing
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/?backend=express
checkout fails as expected

### DIFF
--- a/react/src/components/Products.js
+++ b/react/src/components/Products.js
@@ -2,11 +2,11 @@ import { Component } from 'react';
 import Context from '../utils/context';
 import './products.css';
 import * as Sentry from '@sentry/react';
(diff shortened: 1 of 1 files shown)


ndmanvar (approved):

realkosty:
https://github.com/sentry-demos/empower/pull/271

PR #253, opened 2023-08-15
feat: fill form with default values for checkout

cstavitsky:
- Automatically fill out form with values per https://github.com/sentry-demos/empower/issues/250
- Easter egg: the address is for SF botanical garden

## Screenshot
Screenshot 2023-08-18 at 12 12 10 PM

PR #254, opened 2023-08-16
Format `react/` directory via vscode

cstavitsky:
- Per https://github.com/sentry-demos/empower/pull/253#issuecomment-1679590721
- Formatting changes only! No actual code changes
- Sticking to React directory only so as not to accidentally introduce problems into codebases I don't expect to be touching.


cstavitsky:
@realkosty This is ready for review whenever

realkosty (approved):
Thank you!!

PR #255, opened 2023-08-24
[react] add ?userEmail= query parameter

realkosty:
This is needed to generate some Replay sessions for a particular user. This is needed to demo https://github.com/getsentry/hackweek-zendesk-replay but can have other use cases. Currently all users are generated randomly and we don't have control over them.

# Testing
https://testorg-az.sentry.io/issues/4418158082/?project=4504708328325120&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=0

### DIFF
--- a/react/src/index.js
+++ b/react/src/index.js
@@ -199,41 +199,17 @@ class App extends Component {
 
       scope.setTag('backendType', backendType);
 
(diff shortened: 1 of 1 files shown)


cstavitsky (approved):
👍

PR #266, opened 2023-08-25
test(sourcemaps): explicit empty value

JonasBa:
```
/**
     * One or more paths to ignore during upload.
     * Overrides entries in ignoreFile file.
     *
     * Defaults to `['node_modules']` if neither `ignoreFile` nor `ignore` is set.
     */
    ignore?: string | string[];
```
I wonder if we somehow hit ^ case and we ended up always ignoring node_modules


ndmanvar:
hmmm, interesting. @realkosty - can you check this out when you get a chance?
i.e. can we run locally and verify if "more" sourcemaps are uploaded?

PR #269, opened 2023-08-31
react: add long list render problem

JonasBa:


react/src/components/Products.js
```diff
@@ -66,7 +66,16 @@ class Products extends Component {
     try {
       products = await this.getProducts();
       // take first 4 products because that's all we have img/title/description for
-      this.props.setProducts(products.slice(0, 4));
+      this.props.setProducts(
+        // round to nearest int so we can adjust w/o having to worry about floats
+        Array(Math.round(200 / 4))
```
cstavitsky, line 71: Isn't immediately obvious why we're doing `200/4` rather than `50`. Is this because we have 4 products? If so, consider making a `NUM_PRODUCTS` constant and pulling those references to `4` out into that for readability

cstavitsky (approved):
Looks good, tested locally and saw the long list of products

@JonasBa since there doesn't seem to be a github issue related to this change, can you update description with short description of the expected effect of the change, plus any relevant screenshots (i.e. of sentry transactions or spans), to keep a paper trail?

ndmanvar:
@cstavitsky - can we close this one out?

ndmanvar:
i believe we already introduced a "problem" into the app (in /products)

Issue #276, opened 2023-09-15
[react] uncompressed asset - performance issue


cstavitsky:
Instrument the performance issue: [Uncompressed Asset](https://docs.sentry.io/product/issues/issue-details/performance-issues/uncompressed-asset/) (images, if currently in a GCP bucket, might have to be returned by a server instead)

Luca's thoughts:

> Uncompressed Asset
> Here I just recommend serving a .css file that is very large >1MB on a server that intentionally returns the data very slowly. Additionally the file can’t be compressed and no compression headers can be on the response.
> Since all of these require some kind of external assets to be served in a particular way, maybe it makes sense to have a separate service (uninstrumented) to take care of serving these.

Neil:

> we had `Uncompressed Asset` working at some point.
> Looks like it somehow regressed ): 
> 
> https://github.com/sentry-demos/empower/pull/194
> 
> https://github.com/sentry-demos/empower/blob/master/flask/uc_assets/uc_big_file.js
> 
> https://github.com/sentry-demos/empower/blob/26e230ec5254c9dfec8ae0f5ce68e9c98f0bc2ff/react/src/components/Organization.js#L30-L37

cstavitsky:
- Flask should be hosting the bigfile.js, but for some reason that doesn't exist -- hitting that js file from application-monitoring doesn't produce any results.
- We're running the test via testorganization.py. There's no test code that should be part of this, the test worked at some point. Something is wrong in the application code, whether on the JS or flask side, and next step will be to figure out what

cstavitsky:
This is now working in production per Nar's fix 🥳 

> 1. The event you shared is using an older version of Safari that doesn't support the data attribute we're trying to access to get the size of the asset ([had internal slack conversation about this])
> 2. If we did have that value, there's a typo in the key we're using to extract it, I have a fix pending deployment [here](https://github.com/getsentry/sentry/pull/57233)
> 3. The asset you have downloading contains a query string at the end (?cacheBuster=0.07811220138677588) which is messing with the detection. I have a PR up to fix this [here](https://github.com/getsentry/sentry/pull/57239)

<img width="1792" alt="Screenshot 2023-10-03 at 3 05 20 PM" src="https://github.com/sentry-demos/empower/assets/12092849/617fe985-5a63-4cd2-8d90-0db0a44799d4">

PR #280, opened 2023-09-26
Fix "Uncompressed Asset perf-issue" bug preventing file being fetched from backend

cstavitsky:
Addresses https://github.com/sentry-demos/empower/issues/276

- When trying to get a large file, `uc_big_file.js`, via http://127.0.0.1:8080/uc_assets/uc_big_file.js, the request 404'd.
- Fixed filepaths and renamed for clarity
- 🚧 WIP: still trying to actually trigger the performance issue, although I think it should be triggering [based on the conditions](https://docs.sentry.io/product/issues/issue-details/performance-issues/uncompressed-asset/#:~:text=Uncompressed%20asset%20issues%20happen%20when,that's%20serving%20the%20asset%20file.)


cstavitsky:
Progress update:

Now that the filepath is correct and we can retrieve the file, need to confirm whether the Uncompressed Asset performance issue is triggered from the frontend.

cstavitsky:
@ndmanvar couple questions from when you implemented before:

1. Even when fetching the UC big file successfully, it takes 500ms for the request to complete -- do we expect it to be longer or does this sound right?

Screenshot 2023-09-26 at 4 19 05 PM

2. the UC **small** file 404s, and it does not appear to exist in the `uc_assets/` [directory in flask](https://github.com/sentry-demos/empower/tree/master/flask/uc_assets). I'm not clear on what its purpose is -- should we remove the frontend code that tries to fetch the small file since there is no file to fetch? Or does failing to fetch the small file via a 404 actually do something I'm not understanding?

cstavitsky:
Screenshot 2023-09-26 at 4 37 49 PM

✅ The span has to be one of either: resource.link or resource.script spans, which in practice means it's a .css or .js file.
✅ It has to be over 512kb in size, (to filter out smaller files that are less likely to benefit from being compressed).
✅ The span duration has to be over 500ms (to filter out faster transfers or assets served in internal networks).
❓ There has to be another asset span in the transaction that has compressed files. This helps filter out cases where a users' browser isn't sending the Accept-Encoding header.

"asset span in the transaction that has compressed files." --> Not sure where the line between compressed/uncompressed is, so would need to be sure that the small file I provided is actually compressed..!

cstavitsky:
@ndmanvar mentions that there's a hidden requirement which is to have this occur X times in a certain time frame to start triggering. Won't trigger if you manually hit the page once.

Next steps: merge this in and confirm whether TDA will pick this up

ndmanvar (approved):

PR #282, opened 2023-09-27
upgrade psycopg2-binary

asottile-sentry:
this version supports python3.11


realkosty:
> this version supports python3.11

@asottile-sentry fyi TDA does not work with python3.11 last time I checked. I don't quite remember why and whether that's the case in all environments or only in Google Compute VM or locally. That's why we had to lock python version to 3.8 in virtualenv for TDA:
https://github.com/sentry-demos/empower/blob/6716f2f23a6e13856f35361de082fe8013f434a2/tda/README.md?plain=1#L78

realkosty (approved):

PR #283, opened 2023-09-27
improve flask dev startup by using `flask run`

asottile-sentry:


flask/run.sh
```diff
@@ -5,18 +5,15 @@ set -e
...
     python3 -m venv ./venv
 fi
 source venv/bin/activate
 pip3 install -r requirements.txt
 
 function cleanup {
-  stop.sh python3 $LOCAL_PORT 
-  if [[ "$VIRTUAL_ENV" != "" ]]; then
-    deactivate
-  fi
```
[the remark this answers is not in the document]
realkosty: oh right, I forget that all it does is set env variables
nice video!!

flask/run.sh
```diff
@@ -5,18 +5,15 @@ set -e
...
+  stop.sh python3 $LOCAL_PORT
 }
 trap cleanup EXIT
 
-PORT=$LOCAL_PORT python3 run.py
+flask run --port $LOCAL_PORT
```
realkosty, line 19: should we then also get rid of `run.py` and [this section](https://github.com/sentry-demos/empower/blob/master/flask/src/main.py#L238C11-L244) now that they are no longer used?

asottile-sentry: ah I did delete run.py but I'll delete that too

realkosty (approved):
thank you!

PR #284, opened 2023-09-27
flask: fail on missing environment variables instead of later

asottile-sentry:
prevents some mysterious `NoneType` errors when the environment variables are missing -- now it instead points at exactly what environment variable is needed


realkosty (approved):
Nice!

PR #285, opened 2023-09-27
react: perform all build-related parts in package.json:build

asottile-sentry:
this simplifies calling `npm run build` since it does all the necessary parts so I can use this effectively in a dockerfile

### DIFF
--- a/react/package.json
+++ b/react/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
-    "build": "react-app-rewired build",
+    "build": "INLINE_RUNTIME_CHUNK=false react-app-rewired build && cp serve.json build/",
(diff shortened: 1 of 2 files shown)


react/package.json
```diff
@@ -20,10 +20,16 @@
...
+    "build": "INLINE_RUNTIME_CHUNK=false react-app-rewired build && cp serve.json build/",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
+  "scriptsComments": {
+    "build": [
+      "INLINE_RUNTIME_CHUNK is needed to avoid source map processing error in Sentry when stacktrace includes inline JS",
+      "serve.json is needed to set headers that enable browser profiling"
+    ]
+  },
```
realkosty, line 32: I added this FYI

PR #288, opened 2023-09-29
initial docker-compose setup for local testing

asottile-sentry:
this will eventually be used to run the mock data collection


postgres/data/empowerplant.sql
```diff
@@ -0,0 +1,734 @@
...
+
+
+--
+-- Name: SCHEMA public; Type: ACL; Schema: -; Owner: cloudsqlsuperuser
+--
+
+-- REVOKE ALL ON SCHEMA public FROM cloudsqladmin;
+-- REVOKE ALL ON SCHEMA public FROM PUBLIC;
+-- GRANT ALL ON SCHEMA public TO cloudsqlsuperuser;
+-- GRANT ALL ON SCHEMA public TO PUBLIC;
```
asottile-sentry, line 728: this is the only difference from the `.sqlc`

.github/workflows/mock-data.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches: [master, test-me-*]
+  pull_request:
+
+jobs:
+  mock-data:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - run: docker compose up --wait
+    - run: while ! curl http://localhost:3000; do sleep .5; done
```
asottile-sentry, line 13: there's builtin health checking in docker-compose and docker -- but it's too clunky to use before docker 25 (unreleased)

.github/workflows/mock-data.yml
```diff
@@ -0,0 +1,14 @@
...
+  push:
+    branches: [master, test-me-*]
+  pull_request:
+
+jobs:
+  mock-data:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - run: docker compose up --wait
```
asottile-sentry, line 12: this is pretty slow -- it can be improved if we publish the images to github packages -- will try and do that as a followup

.github/workflows/mock-data.yml
realkosty: should we have a timeout?

asottile-sentry: it's the line right below this one

thinkocapo (commented):
Hey @asottile-sentry @realkosty, reminding that before you mock the data, maybe make sure the data from the apps+services in docker-compose looks correct. Inpsect every error and transaction. I know it's a ton of work, but it's necessary.

Or at least the /products-to-backend tracing and /checkout workflow that we rely on for demo's. Maybe `tracePropagationTargets` will have to get updated.

asottile-sentry:
> Hey @asottile-sentry @realkosty, reminding that before you mock the data, maybe make sure the data from the apps+services in docker-compose looks correct. Inpsect every error and transaction. I know it's a ton of work, but it's necessary.
> 
> Or at least the /products-to-backend tracing and /checkout workflow that we rely on for demo's. Maybe `tracePropagationTargets` will have to get updated.

yeah no I'm not doing that.  the code is identical and the services are wired up the same.  if you want to inspect if by all means check out the branch and look at it but unless you have an actual reason to believe it's different I'm not going to waste my time

thinkocapo:
It's okay @asottile-sentry . I don't have reason to believe it's different, but we often don't, and then stuff still breaks live during demo's 😄 But I think this isn't your focus here, so don't worry about it.

realkosty:
> yeah no I'm not doing that. the code is identical and the services are wired up the same. if you want to inspect if by all means check out the branch and look at it but unless you have an actual reason to believe it's different I'm not going to waste my time

@asottile-sentry I think it will be sufficient that we do that once when everything is working at the high level and all components are in place. I agree this is too much work for incremental PRs. We'd be happy to do that testing ourselves but @thinkocapo is right that it needs to be done before we call the project "done". We have a lot of experience with these demo apps and had seemingly benign minor changes break the demo.

This change (unlike e.g. changes to TDA) also can't break our current set up. If we need extensive changes to TDA at some point I think we should work on it in a forked repo. I do appreciate the self-contained small PRs a lot, it's much easier to digest and very helpful for me to understand the changes. But if you were to merge into a fork (until the eventual integration) maybe that could be more efficient by not having to test for regressions in our current system for every PR.

realkosty (approved):
Have a bunch of comments and suggestions
feel free to ignore or implement in the next iteration
🚢

.github/workflows/mock-data.yml
asottile-sentry: `--fail` is the easier way to do that

react/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /src
+COPY package.json package-lock.json .
+RUN npm install
+
+COPY . .
+
+RUN \
+    REACT_APP_DSN=http://react@localhost:9989/2 \
+    REACT_APP_RELEASE=2023+compose \
+    SENTRY_AUTH_TOKEN=ignored \
```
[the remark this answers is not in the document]
asottile-sentry: yeah I eventually have to reorder this so the build happens outside compose and then the files get stuck into there -- the release / source maps / etc. stuff will get uploaded during that (basically it has to be (1) start fake relay (2) build js (3) build docker images)

react/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    REACT_APP_ASPNETCORE_BACKEND=http://localhost:8091 \
+    REACT_APP_EXPRESS_BACKEND=http://localhost:8088 \
+    REACT_APP_FLASK_BACKEND=http://localhost:8080 \
+    REACT_APP_LARAVEL_BACKEND=http://localhost:8000 \
+    REACT_APP_RUBYONRAILS_BACKEND=http://localhost:5000 \
+    REACT_APP_RUBY_BACKEND=http://localhost:4567 \
+    REACT_APP_SPRINGBOOT_BACKEND=http://localhost:8090 \
+        npm run build
+
+CMD ["serve", "-s", "build"]
```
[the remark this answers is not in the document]
asottile-sentry: the build.sh / run.sh are too specific to macos to be useful (and in many cases are sublty wrong and I didn't want to break them) so I'm ignoring them for now

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+  react:
+    build: react
+    init: true
+    ports: [3000:3000]
+  flask:
+    build: flask
+    environment:
+      <<: *x-base-env
+      FLASK_ENV: test
+      FLASK_APP_DSN: http://flask@mini-relay:9989/3
```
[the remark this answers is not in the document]
asottile-sentry: react is special unfortunately -- it needs to have those variables at *build* time rather than *run* time

Issue #273, opened 2023-09-14
Create a span that causes frontend-only performance slowdown


cstavitsky:
### Current state

In the current demo, we show a slow React transaction whose root cause is directly tied to a slow endpoint that we then trace to a backend service.

- [/products transaction summary](https://demo.sentry.io/performance/summary/?project=5808623&query=transaction.op%3Apageload&referrer=performance-transaction-summary&statsPeriod=30d&transaction=%2Fproducts&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)
- [Example event](https://demo.sentry.io/performance/frontend-javascript:b172523540fd42d0bb3c53b96a0ac51d/?project=5808623&query=transaction.op%3Apageload&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=30d&transaction=%2Fproducts&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)

![Screenshot 2023-09-14 at 1 44 38 PM](https://github.com/sentry-demos/empower/assets/12092849/82fada5b-3501-4e2a-bcd5-a5b004bc3c15)
![Screenshot 2023-09-14 at 1 42 31 PM](https://github.com/sentry-demos/empower/assets/12092849/7ef8c0cd-3997-4e96-a0fc-535c69238add)

This is a strong narrative for customers who are interested in Sentry on the frontend and backend, but doesn’t land as well when speaking with teams of only frontend developers.

### Goal state 

Add a frontend-only slowdown. This new slowdown must:

1. Occur in the frontend-only – not be dependent on performance problems from another downstream service
2. ~Take between 3 and 5 seconds~ (we found this is not possible)
3. ~Jeopardizes/compromises LCP (We want to be able to demonstrate to customers how Sentry surfaces a problem impacting LCP, then permits it to be solved)~ (we are deprioritizing this and making it a separate ticket; the most important thing is that there **is** a frontend-only slowdown that will resonate with a developer, the web vitals are not a critical part of this)
4. Have a specific story/problem associated with it, as well as an indication how a given developer could solve it, based on the insights provided by Sentry. Please make it clear internally what the user would do to solve the slowdown, so that SEs can incorporate this into the demo flow.

##### Ideas/options of how to accomplish goal state

[...]

PR #290, opened 2023-10-05
feat: Allow to pass `?repeat=X` to products/products-join pages

mydea:
With this, you can pass `?repeat=X` to either the `/products` or `/products-join` pages.
This can be used to trigger a slow JS rendering to show frontend performance problems.

Example with 50 repeats: https://sentry-sdks.sentry.io/discover/francescos-test-js-project:ad14e92cf47740298021eda1c0b75346/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&name=All+Events&project=4505997288865792&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29

Example in combination with slow API query: https://sentry-sdks.sentry.io/discover/francescos-test-js-project:d9f34885c0ba41278c61cc541f0019c3/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&name=All+Events&project=4505997288865792&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29

Closes https://github.com/sentry-demos/empower/issues/273

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ localhost:5000?crash=true
 
 Add +2 quantity of a single item to Cart and purchase in order to trigger an Error. Visit the routes defined in src/index.js to produce transactions.
 
+## Trigger slow performance
+
+You can pass `?repeat=X` to the `/products` and `/products-join` URLs, like this: `/products?repeat=100`. 
+This will render a higher number of list items, which shows slow performance in the frontend. 
+You can see a slow react render span in the resulting transaction.
+
 ## Deploy to Prod
 This script deploys the flagship apps React + Flask. For deploying a single app to App Engine, check each platform's README for specific instructions. Make sure you don't have any local changes to `env-config/production.env`.
 ```
--- a/react/src/components/Products.js
+++ b/react/src/components/Products.js
@@ -64,12 +64,16 @@ class Products extends Component {
   async componentDidMount() {
     var products;
     try {
-      products = await this.getProducts();
       // take first 4 products because that's all we have img/title/description for
-      this.props.setProducts(Array(200/4).fill(products.slice(0, 4)).flat().map((p, n) => {
+      products = (await this.getProducts()).slice(0, 4);
+      const repeat = getRepeat() || 50;
+
+      const repeatedProducts = Array(repeat).fill(products).flat().map((p, n) => {
         p.id = n
         return p
-      }));
+      });
+
+      this.props.setProducts(repeatedProducts);
     } catch (err) {
       Sentry.captureException(new Error('app unable to load products: ' + err));
     }
@@ -130,3 +134,9 @@ const mapStateToProps = (state, ownProps) => {
 export default connect(mapStateToProps, { setProducts, addProduct })(
   Sentry.withProfiler(Products, { name: 'Products' })
 );
+
+function getRepeat() {
+  const url = new URL(window.location.href);
+  const repeat = parseInt(url.searchParams.get('repeat'));
+  return Number.isNaN(repeat) ? undefined : repeat;
+}
--- a/react/src/components/ProductsJoin.js
+++ b/react/src/components/ProductsJoin.js
@@ -47,7 +47,14 @@ class ProductsJoin extends Component {
     var products;
     try {
       products = await this.getProductsJoin();
-      this.props.setProducts(products);
+      const repeat = getRepeat() || 1;
+
+      const repeatedProducts = Array(repeat).fill(products).flat().map((p, n) => {
+        p.id = n
+        return p
+      });
+
+      this.props.setProducts(repeatedProducts);
     } catch (err) {
       Sentry.captureException(new Error('app unable to load products'));
     }
@@ -108,3 +115,9 @@ const mapStateToProps = (state, ownProps) => {
 export default connect(mapStateToProps, { setProducts, addProduct })(
   Sentry.withProfiler(ProductsJoin, { name: 'ProductsJoin' })
 );
+
+function getRepeat() {
+  const url = new URL(window.location.href);
+  const repeat = parseInt(url.searchParams.get('repeat'));
+  return Number.isNaN(repeat) ? undefined : repeat;
+}


thinkocapo:
@mydea could you add a note to the README about this feature? So newcomers know it's a possibility. Maybe add it where the README says:
```
localhost:5000?backend=express
localhost:5000/products?se=yourname&backend=express

// throw an exception on any route via 'crash'
localhost:5000?crash=true
```

README.md
```diff
@@ -79,6 +79,12 @@ localhost:5000?crash=true
 
 Add +2 quantity of a single item to Cart and purchase in order to trigger an Error. Visit the routes defined in src/index.js to produce transactions.
 
+## Trigger slow performance
```
mydea, line 82: @thinkocapo added a note here, is that fine you think?

cstavitsky:
@mydea just to keep you updated, I'll be reviewing the PR this week, thanks for your patience (am sick currently so taking longer to get to than I expected)

cstavitsky (commented):
Whew, thanks for the patience here @mydea  😄 

I [deployed to staging](https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/) and tested a bit. I tried loading the page with a variety of values for the repeat query parameter. However, I haven’t managed to get these to meet the full requirements stated in the ticket.

(Definitely let me know if there’s anything I’ve done here while testing that seems off :pray: )

- For reference: example transactions generated for /products-join after deploying your feature branch to [staging](https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/). Events go to [this project](https://demo.sentry.io/projects/staging-frontend-javascript/?project=4504708328325120)
- [/products-join with repeat=100](https://demo.sentry.io/performance/staging-frontend-javascript:1488bef2b8484ad694cf2231e235f2d2/?breakdown=none&project=4504708328325120&query=title%3A%2Fproducts-join&showTransactions=p95&sort=-timestamp&statsPeriod=7d&transaction=%2Fproducts-join)
- [/products-join with repeat=500](https://demo.sentry.io/performance/staging-frontend-javascript:6accafc86ef448ecb77d924e2c091ac6/?breakdown=none&project=4504708328325120&query=title%3A%2Fproducts-join&showTransactions=p95&sort=-timestamp&statsPeriod=7d&transaction=%2Fproducts-join)
- [/products-join with repeat=1000](https://demo.sentry.io/performance/staging-frontend-javascript:51afaa7ced79455bb9df77e8e1d1d83a/?breakdown=none&project=4504708328325120&query=title%3A%2Fproducts-join&showTransactions=p95&sort=-timestamp&statsPeriod=7d&transaction=%2Fproducts-join)


Evaluating the [ticket requirements](https://github.com/sentry-demos/empower/issues/273):

> Add a frontend-only slowdown. This new slowdown must:
> :white_check_mark: Occur in the frontend-only – not be dependent on performance problems from another downstream service

Yep! :+1:

> :white_check_mark: :warning: Take between 3 and 5 seconds

Dependent on the value passed for the repeat query parameter, this condition is met.

However, it seems that once the repeats are bumped up and the slowdown span is bumped up to 3-5 seconds, we regress certain spans being shown. For example, browser assets are no longer shown loading, and this is something we point to in our demo typically.

Browser assets did show at `repeat=100` (see screenshots at bottom of comment) but I'm not sure how consistent/inconsistent this is.

- [existing state in production for /products-join](https://demo.sentry.io/performance/frontend-javascript:5485efbc608e48d5b2fdcb50ae7541de/?project=5808623&query=transaction.op%3Apageload&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=7d&transaction=%2Fproducts-join&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)

Screenshot 2023-10-18 at 11 44 58 AM

- [In staging with `repeat=500`, browser asset spans didn't show](https://demo.sentry.io/performance/staging-frontend-javascript:6accafc86ef448ecb77d924e2c091ac6/?breakdown=none&project=4504708328325120&query=title%3A%2Fproducts-join&showTransactions=p95&sort=-timestamp&statsPeriod=7d&transaction=%2Fproducts-join)

feature_branch_product_join_repeat_500

next requirement --

> :warning: Jeopardizes/compromises LCP (We want to be able to demonstrate to customers how Sentry surfaces a problem impacting LCP, then permits it to be solved)

After testing repeat=100, repeat=200, repeat=500, repeat=1000 on the /products-join page, which represents the frontend-only slowdown, I was not able to see a compromised/jeopardized LCP on any of the transactions. I expected to see the :fire: emoji and red line around the LCP web vital box in a given transaction.

> :warning: Have a specific story/problem associated with it, as well as an indication how a given developer could solve it, based on the insights provided by Sentry. Please make it clear internally what the user would do to solve the slowdown, so that SEs can incorporate this into the demo flow.

1. It would be helpful if you could explain in a few sentences how, if presented with this problem, a developer might realistically solve a problem like this, so SEs know how to complete that story for the devs we are demoing to. Right now I understand that your implementation triggers a slow JS rendering, but I wouldn’t know what to tell a customer in terms of how to solve it.
2. Continuing the story in the product: the natural next step in our demo story, when faced with a 3-5 second slowdown during the transaction, is to leverage frontend browser profiling and examine that 3-5s span more closely via the associated profile.

Previously, viewing the profiling flamegraph would show functions that began around the time of the slow frontend spans.

The behavior I saw while testing does not seem to show relevant function data from a profiling perspective.

@JonasBa - I know this PR modifies some of your recent work to get browser profiling set up in the demo. Do you have an idea of how we might ensure that the profiling story continues to make sense here? This probably should be another ticket but wanted to bring it up awhile...

Screenshot 2023-10-18 at 11 51 42 AM

If this way of thinking about how we would explain profiling to a customer is at all flawed, please let me know.

> (extra) :warning: This was not part of the stated requirements, but I noticed what seems like a potential regression. ui.react.render spans no longer show and we only see ui.react.mount spans. Is this expected?

### Screenshots from testing on staging (same as transactions linked at top of comment)

feature_branch_product_join_repeat_100

feature_branch_product_join_repeat_500

feature_branch_product_join_repeat_1000

mydea:
Hmm, thinking about this some more, I am not sure if it is actually possible to meet all of these goals at the same time.
Trying this out, it seems we flush the page load transaction after some time. so if stuff takes _too_ long (e.g. 3-5 seconds) you'll simply miss some things, including the LCP etc. So I think with the way the SDK works right now, it is actually not possible to do that - because we only wait for a certain time until we end the pageload transaction 🤔 

@AbhiPrasad can you maybe verify that this is correct, or do you have an idea if/how we could otherwise do this?

I also tried adding a huge image and simulating slow network speed. While I was able to slow the page down to a crawl and make it load in >5s, the transaction was still sent at some time below and simply missed these very slow things 😬

cstavitsky:
The original reason I was thinking 3-5 seconds would be a good length is because our transactions that incorporate a backend slowdown [can be quite long, like 15 seconds, and still load browser assets - see screenshot](https://demo.sentry.io/performance/frontend-javascript:550ed67e9fd24122af5e7ee931833704/?project=5808623&query=transaction.op%3Apageload&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=7d&transaction=%2Fproducts&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29). 

3-5 seconds is not a strict requirement, so if that's not possible, it's OK I think. It just needs to be slow enough to feel powerfully bad to a customer in a demo. For instance, if we show a span of 400ms, that may not make the customer think, 'wow, this needs to be fixed asap.'

Screenshot 2023-10-19 at 8 41 55 AM

### Zooming out on what we're trying to do 

It may help to give a high level view of our overall goal, beyond this one ticket. Overall, it's to have the story of a frontend-only slowdown in a transaction that follows the flow below. #4 (and tying in to #5) represents this current ticket.

1. starts with a performance issue (we will most likely trigger the 'uncompressed asset' perf issue)
2. alerts us in slack that there was a frontend performance issue
3. we click in and see the performance issue, which we describe to the customer we're demoing to and explain how Sentry helps us solve
4. after handling the slowdown surfaced by the performance issue, we also notice that the transaction has slow LCP, and see 1+ slow frontend span(s) unrelated to the performance issue **(this ticket is meant to address that)**.
5. from the slow span(s) in #4, we can click through to view a profile that ultimately helps us, in the context of our demo, 'understand' what is going wrong in #4 so we can reduce the transaction duration and improve the LCP value.

(the `/products-join` transaction is currently the best reference transaction for testing things out, because it doesn't contain a backend slowdown)

PR #294, opened 2023-10-17
[tda] run desktop web tests in CI

asottile-sentry:


tda/README.md
```diff
@@ -86,25 +86,6 @@ pip install -r requirements.txt
...
-
-<details>
-<summary>Click to toggle error screenshot</summary>
-
-![Screen Shot 2021-11-29 at 2 57 11 PM](https://user-images.githubusercontent.com/12092849/145083651-5479f05c-107f-4d46-a981-1c728679172f.png)
-
-</details>
-
-**Solution:** A workaround is to locally change the `SAUCELABS_PROTOCOL` constant in `conftest.py` from `https` to `http`.
-
```
asottile-sentry, line 107: this is not something we should let people do -- this is a security issue if we're blindly disabling ssl to a third party

realkosty (approved):

PR #295, opened 2023-10-17
upgrade pyyaml to 6.0.1

asottile-sentry:
6.0 is incompatible with python 3.12


realkosty (approved):

PR #296, opened 2023-10-17
fix datetime.utcnow warning

asottile-sentry:
datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version.


realkosty (approved):

PR #297, opened 2023-10-17
upgrade sentry-sdk to silence warnings

asottile-sentry:
most of these warnings are due to sentry-sdk using `pkg_resources` or `datetime.utcnow(...)`


realkosty (approved):

PR #302, opened 2023-10-19
[tda] security: remove option for saucelabs over http

asottile-sentry:
we should rotate the sauce credentials if someone has sent our creds in plaintext over http


asottile-sentry:
the actual _fix_ here should be in whoever's development environment where they have broken ssl installed -- probably due to a macos upgrade or homebrew upgrade

realkosty (approved):

PR #313, opened 2023-11-03
[tda] code to get latest mobile release from monorepo

realkosty:
# Testing
1. Change code:
```
GITHUB_REPOS = {
    ...
    'ios': 'empower'
}
```
2. Create some test releases in `empower` 
3. Run
```
python3 ./latest_github_release.py ios
ios-0.0.23
```


tda/release_version_manager.py
```diff
@@ -40,6 +51,20 @@ def latest_ios_github_release():
     return latest_github_release('ios')
 
 def determine_latest_release_version(platform):
-    repo_name = GITHUB_REPOS[platform]
-    react_native_releases = requests.get(f"https://api.github.com/repos/sentry-demos/{repo_name}/releases")
-    return react_native_releases.json()[0]['tag_name']
+    repo = GITHUB_REPOS[platform]
+    releases = requests.get(f"https://api.github.com/repos/sentry-demos/{repo['name']}/releases").json()
```
asottile-sentry, line 55: this is probably not going to work or very quickly hit github api limits because it is not authenticated

Issue #308, opened 2023-10-31
Trigger React Error Boundary Linked Error


cstavitsky:
## Goal

The Sentry team is staffing the NY React Summit on Nov 13th. @AbhiPrasad requested to add support for [React Error Boundary Linked Errors](https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/#linked-errors).

## Proposed Approach

Trigger a React Error Boundary Linked Error when the [subscribe button that appears in the footer](https://github.com/sentry-demos/empower/blob/92e9895f60a04b96060829806b5c2be90fe8a254/react/src/components/Footer.js#L15-L31) is clicked. This will be easy for Sentry folks at the conference to demonstrate since we can scroll down from any page, including the homepage, to trigger the error.

![Screenshot 2023-10-31 at 2 05 51 PM](https://github.com/sentry-demos/empower/assets/12092849/45f326f0-ae9e-4a90-9982-916336241392)

cstavitsky:
Note to self for the future: decide if we want to trigger automated data for this specific error

PR #314, opened 2023-11-06
feat(react): Add error boundary to form container

AbhiPrasad:
resolves https://github.com/sentry-demos/empower/issues/308 

Now when you click the subscribe button on the footer for react projects, an error happens and we display a fallback UI

image

This is done via react error boundaries.

image

### DIFF
--- a/react/src/components/Footer.js
+++ b/react/src/components/Footer.js
@@ -1,6 +1,6 @@
 import * as Sentry from '@sentry/react';
-import { Fragment } from 'react';
-import { Component } from 'react';
+import { Fragment, Component } from 'react';
+import { Link } from 'react-router-dom';
 import './footer.css';
 
 class Footer extends Component {
@@ -56,6 +56,9 @@ class Footer extends Component {
               {this.state.subscribed && <SubscribedMessage />}
             </div>
           </Sentry.ErrorBoundary>
+          <p className="sentry-unmask">
+            © 2021 • Empower Plant • <Link to="/about">About us</Link>
+          </p>
         </div>
       </footer>
     );
--- a/react/src/components/Footer.js
+++ b/react/src/components/Footer.js
@@ -13,7 +13,7 @@ class Footer extends Component {
   }
 
   render() {
-    const handleSubscribeClick = () => {
+    const handleSubmit = () => {
       this.setState({ subscribed: true });
     };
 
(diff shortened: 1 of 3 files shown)


react/src/components/Footer.js
```diff
@@ -1,41 +1,69 @@
...
+            onReset={() => {
+              this.setState({ subscribed: false });
+            }}
+            fallback={({ resetError, eventId }) => {
+              return (
+                <Fragment>
+                  <div>An error has occurred. Sentry Event ID: {eventId}</div>
+                  <button className="btn" onClick={resetError}>
+                    Reset Form
+                  </button>
```
AbhiPrasad, line 34: I made a quick call on copy, but we can change to anything else if it fits better.

react/src/components/Footer.js
```diff
@@ -1,34 +1,61 @@
...
-            <input
-              type="submit"
-              value="Subscribe"
-              className="sentry-unmask"
-            ></input>
-          </div>
+                type="submit"
+                value="Subscribe"
+                className="sentry-unmask"
+                onClick={handleSubscribeClick}
```
cstavitsky, line 54: Def not necessary, but consider adding an event handler for hitting the Enter key while on the input. The first time I tested out the new feature I hit Enter to submit the form by reflex, but the page refreshed with no problems instead of triggering the error.

[Example](https://thomasstep.com/blog/react-perform-an-action-when-enter-is-pressed)

cstavitsky (approved):
LGTM 👍 

Left a minor optional suggestion about submitting the form via Enter key. But that's optional and this is good to be merged as-is. @AbhiPrasad if you don't plan to add the Enter key handler, you can merge

✅ tested locally
✅ deployed to staging, tested in staging

## Testing Screenshots
- [link to issue in staging](https://demo.sentry.io/issues/4607309595/?project=4504708328325120&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=0)

Screenshot 2023-11-06 at 2 56 54 PM

Screenshot 2023-11-06 at 2 56 21 PM

react/src/components/Footer.js
AbhiPrasad: done with [d02442f](https://github.com/sentry-demos/empower/pull/314/commits/d02442fc0d3b204de3818f7598e32ab374424162)

PR #335, opened 2023-11-20
[tda] add explicit config setting for collecting mock data

asottile-sentry:


tda/config.local.yaml
```diff
@@ -1,3 +1,5 @@
+collect_mock_data: true
+
```
realkosty, line 2: @asottile-sentry so `dsn` defined on line 7 in this file is the DSN for reporting TDA's own crashes (`empower-tda` project in demo.sentry.io)
Remind me please why are you setting it to mini-relay on localhost? We shouldn't nee to capture those, they are for debugging TDA itself, right?

realkosty (approved):

realkosty:
@asottile-sentry I'm going to go ahead and merge these because my changes depend on these, hope that's okay

PR #336, opened 2023-11-20
[tda] remove pytest.SE_TAG

asottile-sentry:
this is unused, and assigning to pytest globals is discouraged


realkosty (approved):
Oh yeah, this has been vestigial ever since I wrote `RemoteWithExtraUrlParams` eliminating the need to access it from every test
Thanks!

PR #346, opened 2023-11-26
update before_send to read se from event

### LINKED ISSUE #345: Remove configure_scope in flask before_send
https://github.com/sentry-demos/empower/blob/162ea949728de73d0870fdda429ab27dd46f147b/flask/src/main.py#L24-L36

You should never do `configure_scope` in `before_send` because the scope has already been applied at that point. Whatever you want, you should pull from the event  directly in `before_send`.

sdzhong:
Fixes #345

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -27,9 +27,8 @@ RUN_SLOW_PROFILE = None
 def before_send(event, hint):
     # 'se' tag may have been set in app.before_request
     se = None
-    with sentry_sdk.configure_scope() as scope:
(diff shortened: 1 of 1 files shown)


sdzhong:
Validated by running Empower locally with React FE http://localhost:3000/?se=simon which hits Flask BE http://localhost:8080, generating [this event](https://team-se.sentry.io/issues/4672320800/events/69673a6de3f047788a69a76dfa4d0a79/). 
image

✅ It correctly shows `se:simon`

realkosty (approved):

Issue #306, opened 2023-10-20
Make key browser profile frame human-readable


cstavitsky:
> Jonas: There is actually a perfect demo scenario here - you see that we are waiting for server to respond to a request, after which we have a 1s main thread blocked event coming from Products.js component

<img width="692" alt="CleanShot 2023-10-20 at 11 01 37@2x" src="https://github.com/sentry-demos/empower/assets/12092849/6bd822de-fd76-4f9e-aa93-531c9db2436b">

Although profiling works great for this purpose, the `e` frame is not super obvious what it refers to. It would be best to pull the code surfaced by this frame into its own function so that instead of `e` it says something like `renderProducts()` or an otherwise more 'understandable' function name.

> **Jonas Badalic (Fri Oct 20 2023)**
> [@Chris Stavitsky](https://sentry.slack.com/team/U02ACB3981Y)
>  I need to take a look, but if this is some react internal that we cant symbolicate then this will be the best we get
> 🙏1
>
> **Chris Stavitsky**  
> Is there any way we could work around it for the purpose of a clearer demo? (for example, creating an unnecessary function that has a better name; it’s OK if e still is present, but we want something that we can point to that’s human-readable)
>
> **Jonas Badalic**
> I’m not sure, it depends on what this function is and how it’s called
> 👍1 🤞1

Jonas updated:
> Ok, just looked into this - I think the reason this gets lost is because we are using the async keywords here, so the class member methods gets transpiled by babel to a generator function which just so happens to be a great candidate for minification. I need to look more into it, I’m not sure what a solution to this is yet

cstavitsky:
## 11/27/2023 update / thought

Maybe we can use React hooks instead of `componentDidMount`, which lets us get around the `async` usage of `componentDidMount`?

```
import { useState, useEffect } from 'react';

function Proddies() {
  const [productItems, setProductItems] = useState([]);

  useEffect(() => {
    setProductItems([1, 2, 3, 4]);
    alert('hello!');
  }, []);

  return 'Hello world!';
}
export default Proddies;
```

Produces this when locally testing:

![Screenshot 2023-11-27 at 9 47 16 AM](https://github.com/sentry-demos/empower/assets/12092849/fff96b0f-3868-49e1-a811-5af448e980e7)

cstavitsky:
Implemented the above concept in https://github.com/sentry-demos/empower/pull/350 👍

cstavitsky:
[...]

PR #350, opened 2023-11-27
[React] Make browser profile frame human readable

cstavitsky:
- addresses https://github.com/sentry-demos/empower/issues/306
- Replaces the cryptic `e` that is minified due to React limitations, with a human-readable `renderProducts()` function
- tested in staging ✅ 

⚠️ One thing that's not amazing is that we introduce [hooks](https://legacy.reactjs.org/docs/hooks-overview.html) to just one component in order to get this working. Whereas the other components in the app still do not use hooks. Hooks are modern React, and ideally we will rewrite all components in the app to use hooks -- which I wouldn't expect to be that difficult 🤞. But I think it's OK to rewrite components in phases, as general maintenance, and get this PR out there awhile.

## Before
277014355-6bd822de-fd76-4f9e-aa93-531c9db2436b

## After
Screenshot 2023-11-27 at 3 25 19 PM

## Tested on Staging
- [Transaction](https://demo.sentry.io/discover/staging-react:42511f28bc86492abb07d36d9c0d2b9e/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&name=All+Events&project=4504708328325120&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29)
- [Profile](https://demo.sentry.io/profiling/profile/staging-react/9c94716cf1f74dcb9167dc44d7c8d940/flamegraph/?colorCoding=by+system+vs+application+frame&query=&sorting=call+order&spanId=ba34607a72db7e7d&tid=0&view=top+down)
Screenshot 2023-11-27 at 3 43 20 PM

### DIFF
--- a/react/src/components/Products.js
+++ b/react/src/components/Products.js
@@ -10,6 +10,62 @@ import { useState, useEffect } from 'react';
 function Products({ frontendSlowdown, backend }) {
   const [products, setProducts] = useState([]);
 
+  function determineProductsEndpoint() {
+    return frontendSlowdown ? '/products-join' : '/products';
+  }
+
+  function fetchUncompressedAsset() {
+    let uc_small_script = document.createElement('script');
+    uc_small_script.async = false;
+    uc_small_script.src =
+      backend +
+      '/compressed_assets/compressed_small_file.js' +
+      '?cacheBuster=' +
+      Math.random();
+    document.body.appendChild(uc_small_script);
+
+    // big uncompressed file
+    let c_big_script = document.createElement('script');
+    c_big_script.async = false;
+
+    c_big_script.src =
+      backend +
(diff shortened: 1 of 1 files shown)


ndmanvar (approved):

ndmanvar:
@cstavitsky , can you add a comment in changed file on why we cannot use `async` and re. why we think this works

> Jonas: I think what you are seeing is unrelated to hooks and probably related to the async keyword + babel transform, hence why it probably got fixed with hooks (no transform on that class method anymore)

Just to have record of it 9:

ndmanvar (approved):

realkosty:
@cstavitsky this has undone some changes from my PR, fix: https://github.com/sentry-demos/empower/pull/354

PR #359, opened 2023-12-13
[flask] reduce /get_iterator span from 4s to 2s to fudge LCP p75

realkosty:
Make `get_iterator` faster by 2 seconds needed to bring down React LCP p75 down to 5.5s from 7.5s so that the horizontal line in web vitals corresponding to LCP on the right has some thickness to it and looks good in the demo.

Screenshot 2023-12-12 at 10 09 23 PM


realkosty:
@serglom21 @ndmanvar merging without review as this is needed sooner to collect some data

realkosty:
# AFTER
## filtered by `backendType:flask`:
Screenshot 2023-12-12 at 11 24 42 PM
## not filtered 🤨
Screenshot 2023-12-12 at 11 27 04 PM

Fixing this in https://github.com/sentry-demos/empower/pull/360

PR #360, opened 2023-12-13
[flask] make flask backend selection 2x more likely than others

realkosty:
# Motivation
https://github.com/sentry-demos/empower/pull/359 was not enough to get React LCP p75 into the desired range (~ 5.5sec).
Here is why:
Screenshot 2023-12-12 at 11 07 48 PM
This change should fix it.

# Testing 
none


realkosty:
@ndmanvar merging without review as we need data sooner

ndmanvar (approved):

PR #366, opened 2023-12-16
check if tags key is in event's keys

sdzhong:


ndmanvar (approved):

sdzhong:
This resolved the error where there was an unreported KeyError on tags, which causes the event to get dropped & not sent to Sentry. 
This happens when no tags are added to the event, i.e. tags have to be added to the event for it to have a tags key.

The Android and iOS demo apps doesn't include headers for tags (like se or customerType), so the Flask event would not have tags added; thus when checking `if 'se' in event['tags']`, event['tags'] would cause a KeyError. 

With the short-circuit evaluation from this commit (`if 'tags' in event.keys() and 'se' in event['tags']:`), this KeyError is avoided.

PR #368, opened 2023-12-19
Refactor components to use React Hooks

cstavitsky:


react/src/components/Home.js
```diff
@@ -1,20 +1,18 @@
 import { Component } from 'react';
```
sdzhong, line 1: 'Component' is defined but never used. I think this can be deleted.

react/src/components/ButtonLink.js
```diff
@@ -1,14 +1,11 @@
...
 
-class Button extends Component {
-  render() {
-    return (
-      <Link {...this.props} className="btn">
-        {this.props.children}
-      </Link>
-    );
-  }
+function Button({ props }) {
```
sdzhong, line 3: `props` should not be destructured with ({ props }). It should be 
`function Button(props) {`

cstavitsky: addressed in https://github.com/sentry-demos/empower/pull/368/commits/ff645dc059c1f9966870698159a6dd4ea99c9d3d

react/src/components/Home.js
cstavitsky: addressed in https://github.com/sentry-demos/empower/pull/368/commits/ff645dc059c1f9966870698159a6dd4ea99c9d3d

react/src/components/ButtonLink.js
```diff
@@ -1,14 +1,11 @@
...
 
-class Button extends Component {
-  render() {
-    return (
-      <Link {...this.props} className="btn">
-        {this.props.children}
-      </Link>
-    );
-  }
+function Button(props) {
```
sdzhong, line 3: The properties that should be relevant within this component (and for Link) are `to` and `children`. These can be destructured from props and used as needed.

cstavitsky: addressed https://github.com/sentry-demos/empower/pull/368/commits/4a22bbfd9a744a0523ca742b11a4895df889764c

sdzhong (approved):

PR #374, opened 2024-01-12
Refactor react app Components sz

sdzhong:


sdzhong:
Commit https://github.com/sentry-demos/empower/pull/374/commits/fcebf0971bd3ab446babbf9e50fdce74d98828bc is to resolve a couple items:
1. The current use of /nplusone doesn’t capture the http request spans. [Current behavior](https://demo.sentry.io/performance/react:e55fe9241a124edf8a0762c7f868b030/?project=5808623&query=se%3Asimon&referrer=performance-transaction-summary&statsPeriod=24h&transaction=%2Fnplusone&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29); [former behavior](https://demo.sentry.io/performance/react:d1ada29775da434eaa39c260d6ac87f3/) (which was most recently working on Nov 27). There has been no change to [nplusone.js](https://github.com/sentry-demos/empower/blob/master/react/src/components/nplusone.js) in the past 4 months. Maybe an SDK update has affected this.
2. This seems to resolves a Sentry Issue that appears from /nplusone: https://demo.sentry.io/issues/4904022603/events/535b6fdb38534a50ba3c97b82e08ea27/ 
3. The commit now creates the ([perf issue](https://team-se.sentry.io/discover/simon-react:fadc8782f19e4ee8a8cb9fd511706aa7/?dataset=issuePlatform&field=title&field=release&field=environment&field=user.display&field=timestamp&name=N%2B1+API+Call&project=5952962&query=issue%3ASIMON-REACT-2Y&sort=-timestamp&statsPeriod=5m&yAxis=count%28%29)). FYI, I changed the way the page looks, which should be fine (see screenshots of before and after)
image
image

react/src/components/Nav.js
```diff
@@ -9,82 +7,72 @@ import { resetCart, addProduct, setProducts } from '../actions';
 
 import EPlogo from '../assets/empowerplant-logo.svg';
 
-class Nav extends Component {
-  static contextType = Context;
+function Nav(props) {
```
cstavitsky, line 10: Instead of `props`, can we pass the specific props used? At a glance looks like, `cart`, `frontendSlowdown` (I skimmed, but maybe others?)

react/src/components/Products.js
```diff
@@ -7,23 +6,25 @@ import Loader from 'react-loader-spinner';
 import ProductCard from './ProductCard';
 import { useState, useEffect } from 'react';
 
-function Products({ frontendSlowdown, backend }) {
+function Products({ isJoin, frontendSlowdown, backend }) {
```
cstavitsky, line 9: I would prefer we not implement this in this PR since it's a feature change, not a refactor. I don't know whether the broader team will have any concerns about it so I don't want to sneak it into a PR that's meant to keep functionality the same and purely refactor from classes to function components.

(If we do end up doing it, I would also suggest we rewrite things to be `?slowdown=` and then have values `backend` (the backend slowdown), `frontend` (react only slowdown), `none` (join)) 

But my current opinion is to hold off on doing this and run it by others involved in demo like Neil/Kosty

react/src/index.js
```diff
@@ -325,7 +321,7 @@ class App extends Component {
               />
               <Route
                 path="/products-join"
-                element={<ProductsJoin backend={BACKEND_URL} />}
+                element={<Products backend={BACKEND_URL} isJoin={true} />}
```
cstavitsky, line 324: I would suggest we undo this part (see other comment) and undelete ProductsJoin.js for the time being.

cstavitsky (commented):
@sdzhong left a few comments. Nice work!

After comments are addressed, I can re-review and then i'll want to test out locally and/or on staging.

react/src/components/Nav.js
sdzhong: Confirmed that Nav.js was the only one that was missed. Resolved with https://github.com/sentry-demos/empower/pull/374/commits/da23022972a05785fb7ea831cc91f1101ce503ad

react/src/components/Products.js
sdzhong: Understood. I've reverted the changes with https://github.com/sentry-demos/empower/pull/374/commits/6d9b95ac3d881c9a4419ed6d8fcb17b1296827cf.

react/src/index.js
sdzhong: Understood. I've reverted the changes with https://github.com/sentry-demos/empower/pull/374/commits/6d9b95ac3d881c9a4419ed6d8fcb17b1296827cf.

cstavitsky:
Tested locally and can trigger the usual demo flow successfully without issues. 

Deploying to staging, going to let TDA run and check back in a bit or tomorrow to confirm things still look normal traffic wise.

cstavitsky (approved):
I thought TDA ran on staging to generate data but it doesn't, in any case I tested the regular demo error flow on staging and it looks as normal 👍

Issue #367, opened 2023-12-18
Instrument React app with Metrics


serglom21:
Metrics for the JS SDK have been released on [7.88.0](https://github.com/getsentry/sentry-javascript/releases/tag/7.88.0) 🥳 

Could we add the following metrics to our react service please?

- Response Time from backends on /products endpoint request (An average of the response time for all backends) ([get products method](https://github.com/sentry-demos/empower/blob/master/react/src/components/Products.js#L99))-> Distribution
- Conversion Rate for completed purchases ([checkout method](https://github.com/sentry-demos/empower/blob/master/react/src/components/Checkout.js#L41)) -> Increment
- Number of Checkout Errors from FE [(checkout method)](https://github.com/sentry-demos/empower/blob/master/react/src/components/Checkout.js#L41)
- Average order total for successful checkout processes [(access to order total)](https://github.com/sentry-demos/empower/blob/master/react/src/components/Complete.js#L6)

Feel free to add any metrics you see fit or comment or the ones listed above! Thanks!

PR #376, opened 2024-01-19
feat(react): Add custom metrics

ArthurKnaus:
Bump `@sentry/react` to `7.93.0`.
Add custom metrics in react demo.

Screenshot 2024-01-19 at 10 17 43

- closes https://github.com/sentry-demos/empower/issues/367

### DIFF
--- a/react/src/utils/measureRequestDuration.js
+++ b/react/src/utils/measureRequestDuration.js
@@ -1,0 +1,22 @@
+
+import * as Sentry from '@sentry/react';
+
+/**
+ * Measure the duration of a request and send it to Sentry as a custom metric
+ * @param {string} endpoint the endpoint that was called
+ * @returns {() => void} a function to stop the measurement
+ */
+export default function measureRequestDuration(endpoint) {
+  const start = Date.now();
+  
+  function stopMeasurement() {
+    const end = Date.now();
+    const duration = end - start;
+    Sentry.metrics.distribution('request.duration', duration, {
+      unit: 'milliseconds',
+      tags: { endpoint }
+    });
+  }
+
+  return stopMeasurement;
+}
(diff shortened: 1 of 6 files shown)


react/src/utils/measureRequestDuration.js
```diff
@@ -0,0 +1,22 @@
...
+ * @returns {() => void} a function to stop the measurement
+ */
+export default function measureRequestDuration(endpoint) {
+  const start = Date.now();
+  
+  function stopMeasurement() {
+    const end = Date.now();
+    const duration = end - start;
+    Sentry.metrics.distribution('request.duration', duration, {
+      unit: 'milliseconds',
```
serglom21, line 16: @ArthurKnaus The `request.duration` metric does not seem to show any "samples" on the table below ([here](https://demo.sentry.io/ddm/?project=5808623&statsPeriod=7d&widgets=%5B%7B%22mri%22%3A%22d%3Acustom%2Frequest.duration%40milliseconds%22%2C%22op%22%3A%22sum%22%2C%22query%22%3A%22%22%2C%22displayType%22%3A%22line%22%2C%22showSummaryTable%22%3Atrue%2C%22sort%22%3A%7B%22order%22%3A%22asc%22%7D%7D%5D))
Do you know why this is happening? I noticed that here the unit is set to
Screenshot 2024-01-22 at 12 17 27 PM
 `miliseconds` while in the UI, the other distribution metrics have `milisecond` as the unit

ArthurKnaus: @serglom21 good catch! the unit should be `millisecond`, updating...
Regarding the samples:
For now the JS SDK does not support sample correlation. So this is to be expected.

ArthurKnaus:
@serglom21 is anything else missing so I can merge it?

serglom21:
@ArthurKnaus It looks good to me. Thanks!

serglom21 (approved):

PR #379, opened 2024-01-19
add startold for react-scripts start

sdzhong:


sdzhong:
Uploading hot-reload.mp4…

Using `npm run startold` works with hot reload

cstavitsky (commented):
Thanks again for raising this and testing it out with me! I suggest we close this PR in favor of https://github.com/sentry-demos/empower/pull/381, which is now merged -- seems like it's OK to have gotten rid of `react-app-rewired` completely

PR #381, opened 2024-01-19
fix(react): fix cra

JonasBa:
react-app-rewired seems to break HMR so dont use it unless we are bundling the code.


cstavitsky (approved):
LGTM; tested locally and confirmed that hot reload works

JonasBa:
Thank you @cstavitsky!

PR #382, opened 2024-01-23
Make README start instructions clearer

cstavitsky:
Noticed that the `Run` instructions look, at a glance, as if one is supposed to run them sequentially. It would be better to make it extra clear that you pick between two options.


realkosty (approved):
Thanks for fixing 
Screenshot 2024-01-23 at 3 15 18 PM

PR #390, opened 2024-01-30
update python-sdk to 1.40.0 to fix query-source (now on by default)

ndmanvar:
> we fixed the bug with query source not being attached to some spans & we turned collecting query source on by default in [Python SDK 1.40.0](https://github.com/getsentry/sentry-python/releases/tag/1.40.0).

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -75,7 +75,6 @@ class MyFlask(Flask):
             traces_sample_rate=1.0,
             before_send=before_send,
             traces_sampler=traces_sampler,
-            enable_db_query_source=True,
             _experiments={
                 "enable_metrics": True,
                 "profiles_sample_rate": 1.0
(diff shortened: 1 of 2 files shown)


flask/src/main.py
```diff
@@ -75,7 +75,6 @@ def __init__(self, import_name, *args, **kwargs):
             traces_sample_rate=1.0,
             before_send=before_send,
             traces_sampler=traces_sampler,
-            enable_db_query_source=True,
```
realkosty, line 78: is this correct?

ndmanvar: yep. it's defaulted on (:
image
https://github.com/getsentry/sentry-python/releases/tag/1.40.0

serglom21:
Screenshot 2024-01-30 at 6 03 03 PM
Screenshot 2024-01-30 at 6 03 11 PM

Looks good to me! Thanks

serglom21 (approved):

flask/src/main.py
realkosty: ah gotcha, docs didn't get updated it looks

PR #398, opened 2024-02-12
[mini-relay] ignore unclassifiable metric_meta envelopes as well

asottile-sentry:
this should allow us to re-enable the check here

ideally tags would be forwarded along with `metric_meta` and `statsd` and then these could get classified 🤔

### DIFF
--- a/mini-relay/classify_data.py
+++ b/mini-relay/classify_data.py
@@ -45,20 +45,15 @@ class Envelope(NamedTuple):
         except (ValueError, KeyError):
             return None
 
-    def is_session_envelope(self) -> bool:
-        return (
-            self.items[0].header_type in {'session', 'sessions'} and
-            all(
-                item.header_type == 'client_report' for item in self.items[1:]
-            )
-        )
-
-    def is_statsd_envelope(self) -> bool:
-        return (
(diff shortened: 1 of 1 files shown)


mini-relay/classify_data.py
```diff
@@ -45,20 +45,15 @@ def get_trace_id(self) -> str | None:
...
-    def is_statsd_envelope(self) -> bool:
-        return (
-            self.items[0].header_type == 'statsd' and
-            all(
-                item.header_type == 'client_report' for item in self.items[1:]
-            )
+    def is_unclassifiable(self) -> bool:
+        return all(
+            item.header_type in {
+                'session', 'sessions',
```
realkosty, line 51: Having a crash-free-rate is not a critical requirement but nice to have in the long run.
I wonder if we can simulate it without capturing real sessions... after all in `demo` org it's OK if it doesn't match errors.

realkosty (approved):
🚀

mini-relay/classify_data.py
asottile-sentry: yeah I think so -- or at least we can fabricate sessions to match it

PR #414, opened 2024-03-06
Codecov comments added on PR opens/changes

cstavitsky:
### This PR

- When pull requests are opened/re-opened or new commits are pushed in Empower, a Codecov report will be added as a PR comment.

### Evidence

See the [Codecov PR comment](https://github.com/sentry-demos/empower/pull/414#issuecomment-1981758540)!

### What's next

- Open another PR to replace https://github.com/vlad-ko/laravel-stripe-app-gh/pull/42 in Codecov demo
- Add status checks (quality gates)


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          path: empower
+          fetch-depth: 0
+
+      - name: Check out `empower-config` to get env-config
+        uses: actions/checkout@v3
+        with:
+          repository: sentry-demos/empower-config
+          path: empower-config
+          token: ${{ secrets.KOSTY_PERSONAL_ACCESS_TOKEN_FOR_SYNC_DEPLOY_FORK }}
```
cstavitsky, line 19: @realkosty is this `Check out empower-config to get env-config` needed to run tests or no? I wasn't clear, let me know if I should delete it.

cstavitsky:
Demoting to draft - noticed some stuff I think should be different 😳

.github/workflows/codecov.yml
realkosty: Yes, it should not be needed if tests are written properly as true unit tests (e.g. DB connection mocked instead of actually attempting to create a real DB connection with empty host/username/password from missing env variable).

My question is why a separate workflow instead of `auto-deploy.yml`?

realkosty: Oh! this is for comment in PR, sorry never mind
codecov upload in `auto-deploy` is for Sentry code context integration.

cstavitsky: Thanks, will remove it then 👍 

> My question is why a separate workflow instead of `auto-deploy.yml`?

Eventually we want the ability to start a demo from the PR - showcase Codecov comment, then Sentry-bot comment about an error being detected (second part will come later). For this reason we need this workflow to trigger upon PR open/modification within the Empower repo.

The existing auto-deploy workflow currently triggers [upon push to `master`](https://github.com/sentry-demos/empower/blob/master/.github/workflows/auto-deploy.yml#L5-L7). Beyond the trigger being different from what we want, the auto-deploy workflow does a bunch of other things like deployment that we don't want to run when a new PR opens (including deploying).

LMK what you think!

.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -0,0 +1,39 @@
+name: Run Codecov on PR open or change
```
realkosty, line 1:
```suggestion
name: codecov.yml
# Run Codecov on PR open or change
```

.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -0,0 +1,40 @@
...
+          pip install pytz
+          pip install pytest-codecov
+          pytest --cov=. --cov-report=xml
+        working-directory: ./empower/flask/src
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          flags: react, flask
```
cstavitsky, line 38: - Need to upload reports with flags separately. Something like:

```
- name: Upload React coverage reports to Codecov
        uses: codecov/codecov-action@v3
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          flags: react
```

```
- name: Upload Flask coverage reports to Codecov
        uses: codecov/codecov-action@v3
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          flags: flask
```

- Need to add additional config that has something to do with `path` (or may be some other similar name, prob passed under `with` key). Neil suggests it may be `file` or `files` key

cstavitsky: Plan to address this one before merging in 👍

realkosty (approved):
Nice!!

>What's next
>...
>Add status checks (quality gates)

Looks like checks already implemented, no?
Screenshot 2024-04-12 at 2 38 08 PM

cstavitsky:
Oh ya, forgot we have status checks already, leftover from when I started the ticket and forgot to scroll down 😅

PR #417, opened 2024-03-12
[react] Fix grouping by SE "supertag"

realkosty:
After TDA started using the new SE "supertag" format (https://github.com/sentry-demos/empower/pull/332) we're having multiple issues for same flagship error, 1 for each browser/platform combination. This fixes it.

# Testing
> [!NOTE]  
> Tested together with 2 other PRs: https://github.com/sentry-demos/empower/pull/422 and https://github.com/sentry-demos/empower/pull/423

`./deploy.sh --env=local react`
[example event
](https://kosty.sentry.io/issues/5059027877/events/6efcd52f033849f88434fc17b637fef6/)
`./deploy.sh --env=staging react`
https://staging-application-monitoring-react-dot-sales-engineering-sf.appspot.com/products?se=kosty_node_18_webpack_se&backend=rails
[Test error](https://demo.sentry.io/issues/5083997007/events/e65000ef9ea943be9e76731c150afe1f/)
[Test transaction](https://demo.sentry.io/discover/staging-react:cefe18f2a9c842f7adb4951bde23bd22/?field=timestamp&field=event.type&field=transaction&name=Error%3A+500+-+Internal+Server+Error&project=4504708328325120&query=&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29)
[Test error with 'prod-tda-direct-*' se tag to test grouping signature](https://demo.sentry.io/issues/5084021653/events/30e25ea3d46f4c8aad74c2838ee72049/?project=4504708328325120)

PR #422, opened 2024-03-19
[react] Node 16->18, @sentry/webpack-plugin 2.4.0->2.15.0

realkosty:
Google is deprecating Node 16 in App Engine Standard. This fixes the following deployment error:

312173759-8e719220-c0c0-45db-a3a7-8def9dcf94d7

Webpack plugin updated in an attempt to fix sourcemaps in non-production, turned out it was a different issue, kept because why not.

# Testing 
tested together with 2 other PRs in https://github.com/sentry-demos/empower/pull/417


cstavitsky (approved):
LGTM 👍

Issue #424, opened 2024-03-19
Add all query parameter options to README


cstavitsky:
The README should list all query param options:

- `se=myname`
- `frontendSlowdown=true`
- `backend=flask`
- `userFeedback=true`

etc

realkosty:
Thank you :)

cstavitsky:
Completed in https://github.com/sentry-demos/empower/pull/433.

PR #433, opened 2024-03-20
Update react readme with query param options

cstavitsky:
Addresses https://github.com/sentry-demos/empower/issues/424

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -5,3 +5,14 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 ## Deploy
 
 See [../README.md](../README.md)
+
+## Query Parameter Options
+
+(These query parameters can be stacked on top of one another)
+
+- `?backend=flask` - [sets the backend url](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L175-L177) (we have multiple backends with similar functionality). [Here](https://github.com/sentry-demos/empower/blob/master/react/src/utils/backendrouter.js#L4-L10) are all the options for backends).
+- `?userFeedback=true` - displays [(old) user feedback modal](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L208-L212) 5 seconds after checkout crash.
+- `?se=<yourname` - put your name here, for example `?se=chris`, `?se=kosty`, etc. This results in an `se` tag [added to events triggered during the demo](https://github.com/sentry-demos/empower/blob/ea51c3dbce9d50ac32519546e1c772ea5a91722f/react/src/index.js#L191-L198), and [adjust the fingerprint](https://github.com/sentry-demos/empower/blob/ea51c3dbce9d50ac32519546e1c772ea5a91722f/react/src/index.js#L130-L138) to segment issues depending on the SE.
+- `?crash=true` - forces [a crash](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/utils/errors.js#L41)
+- `?userEmail=someemail@example.com` - lets you [pass in a specific user email](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L218-L219)
+- `?frontendSlowdown=true` - used in the [frontend-only demo flow](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L200-L207), which showcases a frontend slowdown via profiling.


cstavitsky:
@realkosty it might be good to also add the sentry.io obscure query param options (I think `issueTracker=jira` or something like that, right, maybe others?) to the demo README somewhere, since it's still demo related -- just not on the side of this application. I don't know if we already have a list of those somewhere 🤔 but we can also do that separately from this

realkosty:
> @realkosty it might be good to also add the sentry.io obscure query param options (I think `issueTracker=jira` or something like that, right, maybe others?) to the demo README somewhere, since it's still demo related -- just not on the side of this application. I don't know if we already have a list of those somewhere 🤔 but we can also do that separately from this

oh good idea

react/README.md
```diff
@@ -5,3 +5,14 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 ## Deploy
 
 See [../README.md](../README.md)
+
+## Query Parameter Options
+
+(These query parameters can be stacked on top of one another)
+
+- `?backend=flask` - [sets the backend url](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L175-L177) (we have multiple backends with similar functionality). [Here](https://github.com/sentry-demos/empower/blob/master/react/src/utils/backendrouter.js#L4-L10) are all the options for backends).
```
realkosty, line 13:
```suggestion
- `?backend=flask` - [sets the backend url](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L175-L177) (we have multiple backends with similar functionality). [Here](https://github.com/sentry-demos/empower/blob/master/react/src/utils/backendrouter.js#L4-L10) are all the options for backends). Additionally recorded in `backendType` tag.  **Default**: flask
```

react/README.md
```diff
@@ -5,3 +5,14 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
...
+- `?userFeedback=true` - displays [(old) user feedback modal](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L208-L212) 5 seconds after checkout crash.
+- `?se=<yourname` - put your name here, for example `?se=chris`, `?se=kosty`, etc. This results in an `se` tag [added to events triggered during the demo](https://github.com/sentry-demos/empower/blob/ea51c3dbce9d50ac32519546e1c772ea5a91722f/react/src/index.js#L191-L198), and [adjust the fingerprint](https://github.com/sentry-demos/empower/blob/ea51c3dbce9d50ac32519546e1c772ea5a91722f/react/src/index.js#L130-L138) to segment issues depending on the SE.
+- `?crash=true` - forces [a crash](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/utils/errors.js#L41)
```
realkosty, line 16:
```suggestion
- `?crash=true` - forces [a crash](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/utils/errors.js#L41) of one of predefined types, selected randomly.
- `?crash=true&errnum=3` - forces crash of specific type depending on `errnum` value
```

react/README.md
```diff
@@ -5,3 +5,14 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
...
+- `?userEmail=someemail@example.com` - lets you [pass in a specific user email](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L218-L219)
+- `?frontendSlowdown=true` - used in the [frontend-only demo flow](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L200-L207), which showcases a frontend slowdown via profiling.
```
realkosty, line 18:
```suggestion
- `?frontendSlowdown=true` - used in the [frontend-only demo flow](https://github.com/sentry-demos/empower/blob/fce289530f72ce47fe2c7482cdbd9aa8bcc13b6e/react/src/index.js#L200-L207), which showcases a frontend slowdown via profiling.
```

> [!NOTE]  
> While not an Empower React app query param, we also have demo-specific query params to customize Sentry UI experience. Those have to be set in the Sentry app itself.
> - `&issueTracking=jira` - only shows 1 issue integration instead of all ~10 ([internal documentation](https://www.notion.so/sentry/issueTracking-hack-c8bbae41bcb84c80aed9f3fc0ab29df6))

realkosty (changes requested):
Thank you for documenting all these. Minor tweaks and `issueTracking` as you requested.

Also should we delete [Port Notes section](https://github.com/sentry-demos/empower/blob/query_param_readme/README.md#port-notes) as this duplicates it?

cstavitsky:
> Also should we delete [Port Notes section](https://github.com/sentry-demos/empower/blob/query_param_readme/README.md#port-notes) as this duplicates it?

@realkosty agree, deleted that section. I also moved the query param stuff to the main README (i'd had it in the `react/` README but on more thought figured nobody will easily find that, and query params should be easily discoverable

realkosty:
> > Also should we delete [Port Notes section](https://github.com/sentry-demos/empower/blob/query_param_readme/README.md#port-notes) as this duplicates it?
> 
> @realkosty agree, deleted that section. I also moved the query param stuff to the main README (i'd had it in the `react/` README but on more thought figured nobody will easily find that, and query params should be easily discoverable

oh good call!

realkosty (approved):
Love it.
might even want to move query params section to the top as more people will be using the app vs contributing (Setup, Run)

Issue #257, opened 2023-08-24
[react] rage clicks, dead clicks for Session Replay


cstavitsky:
[Blog post](https://blog.sentry.io/introducing-rage-and-dead-click-detection-for-session-replay/)

Related: https://github.com/sentry-demos/empower/pull/256 (this permits rage/dead clicks, but as far as I know we haven't opened anything to change our automated jobs to actually create scenarios where rage/dead clicks will happen)

- Automated data should trigger rage clicks
- Automated data should trigger dead clicks

realkosty:
> we haven't opened anything to change our automated jobs to actually create scenarios where rage/dead clicks will happen
we have not afaik

PR #447, opened 2024-03-26
Trigger rage click after checkout error

cstavitsky:
- Addresses https://github.com/sentry-demos/empower/issues/257
- Adds TDA test to trigger rage clicks as a part of checkout flow
- Can be triggered manually by clicking button "Contact Us" after the checkout error (click 5+ times, wait 7+ seconds to trigger)
- Can also be triggered instead of checkout flow by passing query param `?rageclick=true`, which disables the checkout button functionality, and then 'rageclicking' the checkout button.
- Generated[ rage click from running TDA locally](https://team-se.sentry.io/issues/5112756355/?end=2024-03-28T23%3A03%3A24&project=4506984343273472&query=se%3Acstavitsky-tda-direct-master%2A+%21project%3Aempower-tda&referrer=issue-stream&start=2024-03-28T23%3A02%3A12&stream_index=1/)

# Screenshots
Screenshot 2024-03-27 at 5 09 47 PM

Screenshot 2024-03-27 at 5 08 55 PM


tda/desktop_web/test_checkout.py
```diff
@@ -57,11 +61,14 @@ def test_checkout(desktop_web_driver, endpoints, batch_size, backend, random, sl
                 desktop_web_driver.find_element(By.CSS_SELECTOR, 'a[href="/checkout"]').click()
                 time.sleep(sleep_length())
 
-                desktop_web_driver.find_element(By.CSS_SELECTOR, '#email').send_keys("sampleEmail@email.com")
```
realkosty, line 60: I understand this is no longer necessary since you changed the app to pre-fill checkout address etc, right?
we never ended up changing that behavior depending on `se`?

realkosty (approved):
I love that this is a contact us button on `/error` page. Double whammy. I can already see how this is going to make our demo even more entertaining and emotionally resonant.

tda/desktop_web/test_checkout.py
cstavitsky: > I understand this is no longer necessary since you changed the app to pre-fill checkout address etc, right? 

Yep - when i was watching sauce video, i noticed it double filling email and the validation complaining. I'm not sure whether that caused any problems for sauce runs but made sense to remove it

> we never ended up changing that behavior depending on `se`?

Could you clarify what you mean here? I don't understand

cstavitsky:
Couple updates:

- Updated functionality to allow `?rageclick=true` query param per @ndmanvar suggestion, so that the 'rage click' can take the place of the error flow
- TDA will generate rage clicks in this way via a separate test from checkout
- Still kept the 'contact us' button so that SEs can manually trigger a rageclick after showing the checkout error

tda/desktop_web/test_checkout.py
realkosty: I don't remember the details (probably happened on slack) but vaguely remember that initial change (https://github.com/sentry-demos/empower/pull/253) broke TDA and was later fixed with https://github.com/sentry-demos/empower/commit/3493cfead5b0803d5614b08f712779c69044377f

ndmanvar (approved):

tda/desktop_web/test_rageclick.py
```diff
@@ -0,0 +1,86 @@
...
+                sentry_sdk.set_tag("missedButtons", missedButtons)
+
+                if err:
+                    sentry_sdk.capture_exception(err)
+
+            time.sleep(sleep_length())
+
+
+
+
```
ndmanvar, line 81: remove whitespace @cstavitsky

ndmanvar (approved):

PR #448, opened 2024-03-27
Add react component names

c298lee:
Adds react component names: https://docs.sentry.io/platforms/javascript/guides/react/features/component-tracking/
React component names only works on jsx and tsx files, so I changed all component files from js to jsx


cstavitsky:
@c298lee can you confirm this is the expected outcome? Based on the docs ("Now, you will be able to see your component names in breadcrumbs, spans, and search for replays by component names"), seems like it is to me, but would be great if you can confirm.

Screenshot 2024-03-28 at 4 37 40 PM
Screenshot 2024-03-28 at 4 37 14 PM

cstavitsky (approved):
LGTM 👍 just left a quick comment confirming expected behavior from my local testing

c298lee:
@cstavitsky Yes that is the expected outcome!

PR #455, opened 2024-04-24
remove "Permit your IP address in CloudSQL" from README.md

realkosty:
IP whitelisting for database is no longer necessary


realkosty:
already part of current version

PR #463, opened 2024-05-03
Add logic to fetch subscription plans

cstavitsky:
- Adding a function and a test

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -24,7 +24,7 @@ github_checks:
   annotations: true
 
 comment:
-  layout: "newheader, diff, flags, files"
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -37,19 +37,15 @@ flag_management:
...
     - name: frontend
       paths:
         - react/
-      statuses:
-        - type: project
-          target: auto
-          threshold: 1%
   default_rules:
     carryforward: true
+    statuses:
```
cstavitsky, line 45: Let's move this out of this PR -- make a separate PR and merge to `main`, so this change will go away. Want to only show 'real code changes' in this PR.

PR #471, opened 2024-05-09
switch dummy emails from gmail to example.com

cstavitsky:
I didn't test this since it's such a simple change, but what could go wrong, right? Right? 😅 

I know we're masking them in sentry, but there are still opportunities for this to be problematic. Better to just fix it at the source. Example 'problem' is this showing up:

Screenshot 2024-05-08 at 4 03 04 PM

Sometimes marketing or docs needs to take screenshots, so might as well just use example.com everywhere


realkosty (approved):
TY

PR #513, opened 2024-05-15
LaunchDarkly

sdzhong:
Add LaunchDarkly feature flag for [LaunchDarkly's Sentry Integration](https://docs.launchdarkly.com/integrations/sentry).

### DIFF
--- a/react/src/components/About.jsx
+++ b/react/src/components/About.jsx
@@ -59,24 +59,23 @@ function About({ backend }) {
     if(lDContext.lDClient) {
       const ldClient = lDContext.lDClient;
       const ldContext = ldClient.getContext();
-      Sentry.setContext('launchdarklyContext', {
-        "kind": ldContext.kind,
-        "key": ldContext.key,
-      });
-
       // throw error if LaunchDarkly Feature Flag is enabled
       const ldFlag = 'my-sentry-integration-feature';
       if (ldClient.variation(ldFlag, false)) {
         console.log(`LaunchDarkly flag ${ldFlag} is enabled!`);
-        Sentry.configureScope(function (scope) {
-          Sentry.setContext('response', request1);
+        Sentry.withScope(function (scope) {
+          scope.setContext('response', request1);
+          scope.setContext('launchdarklyContext', {
+            "kind": ldContext.kind,
+            "key": ldContext.key,
+          });
+          Sentry.captureException(
+            new Error(
+              request1.status + ' - Error from Feature Flag (LaunchDarkly)'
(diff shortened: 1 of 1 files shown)


sdzhong:
fb566048704724865284e1a545a47d161cdf868b _convert index.js class component to functional component_ is because I want to use hooks (useState + createContext) to fetch LaunchDarkly properties in index.js then propagate to child components.

react/src/index.js
```diff
@@ -285,7 +289,36 @@ const App = () => {
   // Crasher parses query params sent by /tests for triggering crashes for Release Health
   crasher();
 
+
+  //LaunchDarkly settings
+  const [lDFeatureFlag, setLDFeatureFlag] = React.useState(false);
+  const lDUser = {
+    "kind": queryParams.get('se') === null || !queryParams.get('se').match("-tda-") ? "user": "automation",
```
sdzhong, line 296: If the se tag is not empty and contains _-tda-_, e.g. prod-tda-direct--desktop_web/test_products_join.py-sauce-chrome_windows, then user > kind: automation
Otherwise, user > kind: "user"

sdzhong: My idea is that if kind: automation, then it would be part of the [Experiment](https://docs.launchdarkly.com/home/experimentation); if kind: user, then it would not.

react/src/components/About.jsx
```diff
@@ -56,24 +56,34 @@ function About({ backend }) {
...
-        Sentry.captureException(
-          new Error(
-            r.status + ' - Error from Feature Flag (LaunchDarkly)'
-          )
-        );
-        console.log(`Generated error specific to LaunchDarkly feature flag`);
+    if(lDContext.lDClient) {
+      const ldClient = lDContext.lDClient
+      ldContext = ldClient.getContext()
+      Sentry.setContext('launchdarklyContext', {
```
sdzhong, line 62: This doesn't appear to work. Example Sentry `Error from Feature Flag (LaunchDarkly)` Issue event context: 
image


Here is an example of what it looked [before](https://github.com/sentry-demos/empower/pull/513/commits/b552ea069704bd46dc333ea8966656f030f965b6):
image

sdzhong: @mike-zorn , can you please take a look? Is 
```
"kind": ldContext.kind,
"key": ldContext.key,
```
valid?

sdzhong: I can't seem to reproduce this. I'll reopen if needed but this might be a false alarm.

PR #525, opened 2024-06-04
Bump Python SDK version

sentrivana:
- Bumped Python SDK to 2.4.0. There have been no breaking changes in the API that we're using in Empower, so no code changes necessary.
- Removed one unused import.

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -10,7 +10,6 @@ import dotenv
 from .db import get_products, get_products_join, get_inventory
 from .utils import parseHeaders, get_iterator
 import sentry_sdk
-from sentry_sdk import metrics
 from sentry_sdk.integrations.flask import FlaskIntegration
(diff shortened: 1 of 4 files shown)


flask/src/main.py
```diff
@@ -10,7 +10,6 @@
 from .db import get_products, get_products_join, get_inventory
 from .utils import parseHeaders, get_iterator
 import sentry_sdk
-from sentry_sdk import metrics
```
sentrivana, line 13: We're emitting metrics in this file, but via `sentry_sdk.metrics`.

realkosty:
@sentrivana thank you! I'll test for regressions this week (lmk if this is a blocker and we need it sooner)

sentrivana:
Thank you @realkosty! Not a blocker, just updating the SDK across Sentry so that we're eventually on 2.0+ everywhere.

realkosty (approved):

PR #526, opened 2024-06-05
Bump .NET SDK

bitsandfoxes:
Bumped the SDK and updated the `User` to `SentryUser` due to a breaking change in the mayor bump.

### DIFF
--- a/aspnetcore/Empower.Backend.csproj
+++ b/aspnetcore/Empower.Backend.csproj
@@ -9,12 +9,12 @@
     <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
   </PropertyGroup>
   
-<!--  <PropertyGroup>-->
-<!--    <SentryOrg>$(SENTRY_ORG)</SentryOrg>-->
-<!--    <SentryProject>$(ASPNETCORE_SENTRY_PROJECT)</SentryProject>-->
-<!--    <SentryUploadSymbols>true</SentryUploadSymbols>-->
-<!--    <SentryUploadSources>true</SentryUploadSources>-->
-<!--  </PropertyGroup>-->
+  <PropertyGroup>
+    <SentryOrg>$(SENTRY_ORG)</SentryOrg>
+    <SentryProject>$(ASPNETCORE_SENTRY_PROJECT)</SentryProject>
+    <SentryUploadSymbols>true</SentryUploadSymbols>
+    <SentryUploadSources>true</SentryUploadSources>
+  </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="7.0.4" />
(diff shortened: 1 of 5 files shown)


aspnetcore/Empower.Backend.csproj
```diff
@@ -16,9 +16,13 @@
     <SentryUploadSources>true</SentryUploadSources>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+    <UseSentryCLI>false</UseSentryCLI>
+  </PropertyGroup>
```
realkosty, line 21: hm why not leave the warning?

realkosty (changes requested):
Thank you for the update!!
just 1 comment

aspnetcore/Empower.Backend.csproj
bitsandfoxes: There is no `sentry-cli` auth stuff on the environment when it gets built by CodeQL and then fails horribly, breaking the build.

bitsandfoxes: So we'd either need to find a way to set those and the auth token somehow, somewhere to be picked up by the action, which I don't know how to do, or we skip the upload for `debug` builds (for now (tm)).

realkosty: Hm chatgpt's telling me we could set a condition like this
<PropertyGroup Condition="'$(CODEQL_ANALYSIS)' == 'true'">
    <UseSentryCLI>false</UseSentryCLI>
</PropertyGroup>

realkosty (approved):

aspnetcore/Empower.Backend.csproj
realkosty: Is that the same failure you had before when CodeQL was failing due to missing auth token?
 ```
 Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
  - Required package "facade/ignition" is not present in the lock file.
  - Required package "nunomaduro/collision" is not present in the lock file.
  This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.
  Read more about correctly resolving merge conflicts https://getcomposer.org/doc/articles/resolving-merge-conflicts.md
  and prefer using the "require" command over editing the composer.json file directly https://getcomposer.org/doc/03-cli.md#require-r
  Error: Could not auto-detect a suitable build method
  Error: We were unable to automatically build your code. Please replace the call to the autobuild action with your custom build steps. Encountered a fatal error while running "/opt/hostedtoolcache/CodeQL/2.17.4/x64/codeql/csharp/tools/autobuild.sh". Exit code was 1 and last log line was: Error: Could not auto-detect a suitable build method. See the logs for more details.
```

realkosty: Either way, we can figure it out later 🚢

bitsandfoxes: Looks like ChatGPT is dreaming things up as they should be and not how they are 😅

Issue #530, opened 2024-06-18
Update to v8 of the JavaScript SDK


mydea:
We need to update the demo to use v8 of the SDK. There is some manual instrumentation that we'll need to convert to the new performance APIs.

PR #532, opened 2024-07-04
Upgrade react sdk to v8

andreiborza:
Closes: #530


react/src/components/About.jsx
```diff
@@ -41,9 +41,7 @@ function About({ backend }) {
     // Error Handling
     response.forEach((r) => {
       if (!r.ok) {
-        Sentry.configureScope(function (scope) {
-          Sentry.setContext('response', r);
-        });
+        Sentry.getCurrentScope().setContext('response', r);
```
mydea, line 44: let's instead make this:

```js
Sentry.withScope((scope) => {
  scope.setContext(...);
  Sentry.captureException(...);
});
```

This is the preferred pattern over using current scope like this!

react/src/components/Products.jsx
```diff
@@ -103,11 +103,9 @@ function Products({ frontendSlowdown, backend }) {
       })
         .then((result) => {
           if (!result.ok) {
-            Sentry.configureScope(function (scope) {
-              Sentry.setContext('err', {
-                status: result.status,
-                statusText: result.statusText,
-              });
+            Sentry.getCurrentScope().setContext('err', {
```
mydea, line 106: let's use `Sentry.setContext(..)` instead

react/src/components/ProductsJoin.jsx
```diff
@@ -16,11 +16,9 @@ function ProductsJoin({ backend }) {
     })
       .then((result) => {
         if (!result.ok) {
-          Sentry.configureScope(function (scope) {
-            Sentry.setContext('err', {
-              status: result.status,
-              statusText: result.statusText,
-            });
+          Sentry.getCurrentScope().setContext('err', {
```
mydea, line 19: let's use `Sentry.setContext(..)` instead

react/src/index.js
```diff
@@ -72,6 +72,8 @@ Sentry.init({
   release: RELEASE,
   environment: ENVIRONMENT,
   tracesSampleRate: 1.0,
+  tracingOrigins: tracingOrigins,
```
mydea, line 75:
```suggestion
```

react/src/index.js
```diff
@@ -192,86 +165,85 @@ class App extends Component {
...
-        RAGECLICK = true;
-      }
+    const scope = Sentry.getCurrentScope();
+    const customerType = [
+      'medium-plan',
+      'large-plan',
+      'small-plan',
+      'enterprise',
+    ][Math.floor(Math.random() * 4)];
+    scope.setTag('customerType', customerType);
```
mydea, line 175: let's not use current scope here, and just do `Sentry.setTag(...)` etc!

mydea (commented):
generally looks good, mostly I'd replace usage of `getCurrentScope()` with just using top level methods!

react/src/components/Products.jsx
andreiborza: Thanks, updated.

react/src/components/ProductsJoin.jsx
andreiborza: Updated.

react/src/index.js
andreiborza: Updated and here and the other places that used this scope.

react/package.json
```diff
@@ -19,7 +19,7 @@
     "web-vitals": "~1.0.1"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "react-app-rewired start",
```
andreiborza, line 22: This breaks HMR, see https://github.com/sentry-demos/empower/pull/381

Need to rewire jsx-runtime but also want HMR... 🤔

andreiborza: HMR was already broken, but I reverted this change anyway...

mydea (approved):
LGTM! 🚀

react/src/index.js
```diff
@@ -80,44 +81,15 @@ Sentry.init({
...
-      tracingOrigins: tracingOrigins,
-      tracePropagationTargets: tracingOrigins,
-      routingInstrumentation: Sentry.reactRouterV6Instrumentation(
-        useEffect,
-        useLocation,
-        useNavigationType,
-        createRoutesFromChildren,
-        matchRoutes
-      ),
-      beforeNavigate: (context) => {
```
cstavitsky, line 95: @andreiborza to confirm, is the reason for removing the beforeNavigate block that it's outdated and no longer necessary?

andreiborza: yes, the newer `reactRouterV6BrowserTracingIntegration` doesn't have an option for that and from what I can tell, it's (just?) doing route parametrization which the newer integration does out of the box.

Could you also chime in here please @mydea?

cstavitsky (approved):

react/src/index.js
mydea: sorry, saw this too late! This should not be needed here, if needed you can use `beforeStartSpan` instead, see https://docs.sentry.io/platforms/javascript/tracing/instrumentation/automatic-instrumentation/#beforestartspan

PR #537, opened 2024-07-26
Pretttty sure we can delete this variable...

realkosty:
### very small change, no need to test!


realkosty:
### I'm just gonna go ahead and merge it without approval, it's such a simple change

PR #541, opened 2024-07-29
Update Checkout.jsx

drguthals:
Updating the email address and mailing address to be actual examples instead of potential real ones.


ndmanvar:
@cstavitsky - can you review/approve/merge when you get a chance? 🙏

cstavitsky (approved):
Thanks for thinking of these @drguthals ✌️

PR #544, opened 2024-08-08
Add tests for cc coverage

ndmanvar:


ndmanvar:
@cstavitsky This is ready for review.
I got it to ~40%

The story will be we have ~40% code coverage right now, and we are enforcing patches are 80% tested.
Therefore, we can increase the coverage over time (so we can get to 70% eventually)

React has a bunch of tests that all work/pass.
Flask has 3 test files (`test_utils.py`, `test_db.py`, `test_main.py`).
`test_db.py` and `test_main.py` are commented out, as I couldn't finish those. Keeping those checked-in so that we can revisit if needed.

cstavitsky (approved):
<didn't read every line, just skimmed but looks like we're changing only tests and codecov is picking up accurately! Awesome 🎉 >

PR #551, opened 2024-08-21
[mock-data] collect test data... slowly

asottile-sentry:
depends on #550


mini-relay/run_tests.py
```diff
@@ -0,0 +1,145 @@
...
+        cmd,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    if ret:
+        raise SystemExit(f'`{shlex.join(cmd)}` raised {ret}')
+
+
+def _discover_tests(args: list[str]) -> list[str]:
+    cmd = ('pytest', 'tda/desktop_web', '--collect-only', '--quiet', *args)
```
realkosty, line 40: nice

mini-relay/run_tests.py
```diff
@@ -0,0 +1,145 @@
...
+        _wait_for_started('http://localhost:3000')
+    except BaseException:
+        _run_q('docker', 'compose', 'down')
+        raise
+
+    try:
+        yield
+    finally:
+        print('... waiting for eventual consistency')
+        time.sleep(10)
```
realkosty, line 99: it bothers me a bit that we haven't figured out why some envelopes come with delay

mini-relay/run_tests.py
```diff
@@ -0,0 +1,145 @@
...
+            # processes are much more likely to bind SIGINT than SIGTERM
+            '--signal=SIGINT',
+            # specifically leaving postgres / mini-relay running
+            'react', 'ruby', 'flask',
+        )
+        time.sleep(2)
+        _run_q('docker', 'compose', 'down')
+
+        testpart = testname.rpartition('::')[-1]
+        testpart = testpart.partition('[')[0]
```
realkosty, line 112: note to self: verify we fixed test filenames not matching test function names

mini-relay/run_tests.py
```diff
@@ -0,0 +1,145 @@
...
+
+        print('... saving mock data')
+        dt = datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
+        target = os.path.join('mini-relay', 'classified', f'{dt}_{testpart}')
```
realkosty, line 116: could potentially use commit hash instead or in addition to `dt` 
just intuition - not sure I have a good sense of how exactly we'll be using the filename down the line

realkosty (approved):
🚀

mini-relay/run_tests.py
asottile-sentry: it kinda makes sense -- sending everything immediately is expensive and if you can batch stuff better then you save on network overhead

realkosty: do you think that's happening at HTTP level or our SDKs transport?

asottile-sentry: SDKs themselves -- I know the python one at least does some amount of buffering / flushing for performance

PR #554, opened 2024-08-27
fix replay .sentry-unblock after v7->v8 everything is ******

realkosty:
https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/#change-of-replay-default-options-unblock-and-unmask

Let's actually follow migration docs next time 😄

BEFORE
Screenshot 2024-08-26 at 5 40 07 PM

AFTER
Screenshot 2024-08-26 at 5 40 35 PM


realkosty:
Also just noticing "Report a bug" is redacted

cstavitsky (approved):

PR #576, opened 2024-09-25
[react] Don't fail deploy if a test fails

realkosty:
IMO we should not have to maintain fake unit tests unless this is important for Codecov demo.

Context: https://github.com/sentry-demos/empower/pull/577

### DIFF
--- a/.github/workflows/auto-deploy.yml
+++ b/.github/workflows/auto-deploy.yml
@@ -69,10 +69,10 @@ jobs:
           echo "BASE: ${{ env.NX_BASE }}"
           echo "HEAD: ${{ env.NX_HEAD }}"
 
-      - name: Run React tests
+      - name: Run React tests (not failing deployment if a test fails)
         run: |
           npm install
-          npm test -- --coverage
+          npm test -- --coverage || true
         working-directory: ./empower/react
       - uses: codecov/codecov-action@v3
         with:


cstavitsky:
@realkosty, hmm does this mean the `codecov.yml` check is going to fail on every single PR?

Does this 'false negative', which i can see us starting to ignore on every run, risk obscuring actual future problems occurring in the Codecov test run? 🤔 

Screenshot 2024-09-25 at 9 52 50 AM

That said, i understand not wanting to maintain fake unit tests... let's talk more about this

realkosty:
> @realkosty, hmm does this mean the `codecov.yml` check is going to fail on every single PR?

yes, after the first time some existing test starts failing

> Does this 'false negative', which i can see us starting to ignore on every run, risk obscuring actual future problems occurring in the Codecov test run? 🤔

yes. The way it's implemented if test run command `npm test -- --coverage` fails for ANY reason, deployment would still succeed. 

We could add a check that coverage report is generated after this step to mitigate that.

realkosty:
@cstavitsky Maybe this issue won't come up very often and not worth our time?
Ideally we probably want to have tests for code that doesn't get changed a lot
What's the benefit of having our codecov demo use the same repo as React instead of some random sample repo btw?

cstavitsky:
> What's the benefit of having our codecov demo use the same repo as React instead of some random sample repo btw?

IIRC the vision was to incorporate other codecov features into the PR and possibly have the PR serve as an entry point into the sentry demo as well (part of the platform vision)

Although I'm not sure how much overlap there has been between a Codecov POC and Sentry POC for the same customer

We should discuss with @ndmanvar

yuval-sentry (approved):
LGTM

realkosty:
Discussed offline, decided we will just change all assertions to `assert(true)` instead of this change

PR #577, opened 2024-09-25
[react] fix unit test broken by Crit. Exp. PR and blocking deploy

realkosty:
Was breaking auto-deploy https://github.com/sentry-demos/empower/actions/runs/11023851879/job/30615935888


realkosty:
self-approving to unblock deployment

PR #587, opened 2024-09-30
Update js/react sdk to 8.32.0

ndmanvar:


realkosty (approved):
spoke with Neil - he checked release notes, should not be any breaking changes

PR #599, opened 2024-10-22
Exclude Next from auto-deploy since it has it's own Vercel automation

realkosty:
currently breaking our pipeline https://github.com/sentry-demos/empower/actions/runs/11432793019/job/31803804101

I wonder if Vercel's bot will now post in all commits in this repo (this one modified file in root directory, not next).. maybe there's a way to limit it to changes in `next` directory only? I'd say it's not a big deal except that it adds noise to codecov demo. @ndmanvar what do you think?

### DIFF
--- a/auto-deploy.exclude
+++ b/auto-deploy.exclude
@@ -8,3 +8,5 @@ postgres
 crons-python
 ruby-on-rails 
 # ^ see https://github.com/sentry-demos/empower/issues/439
+next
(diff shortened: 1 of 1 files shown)


landenai:
Screenshot 2024-10-22 at 10 04 04 AM
fyi i enabled this setting to only deploy on changes to the root directory (next) yesterday afternoon

landenai (approved):

PR #608, opened 2024-10-25
Add LLM plant suggestions and monitoring based on city and weather

lukemun:
* Created a new button "Get Suggestion" on plants homepage
* Created flask endpoint /suggestion
* Fetches a suggestion based on weather and users city
* Instruments llm monitoring in sentry


realkosty (commented):
beautiful.

for this to work in production we'll need to figure out where to put OPENAI_API_KEY
normally we add things like that to this private repo https://github.com/sentry-demos/empower-config/blob/main/production.env (prefixed with REACT_/FLASK_/etc) but since this is a real key (as opposed to password to fake database), security will not love this even though it's a private repo. 

Is there a way we can fake this without making real calls to OpenAI?

lukemun:
Too many conflicts, was easier to just create a new [branch](https://github.com/sentry-demos/empower/pull/625)

PR #616, opened 2024-10-31
Moved components and fixed loading spinner

landenai:
- consolidated reusable components into `components` dir
- fixed loading spinner


next/src/components/ThreeDotLoader.jsx
```diff
@@ -1,11 +1,13 @@
 'use client'
 
-import Loader from 'react-loader-spinner';
```
lukemun, line 3: Could we keep this 3 dot loader in the /ui directory? General purpose reusable components should live there and custom made ones for a particular page should be in /components

PR #625, opened 2024-11-12
LLM Suggestion local runs

lukemun:
- Adds LLM feature to next
- Creates endpoints in flask
- Adds conditional logic checking for OPEN_API_KEY
- Update READEME.md


README.md
```diff
@@ -175,3 +175,14 @@ gcloud config list, to display current account
...
 
 `gcloud app deploy` does not support `--update-env-vars RELEASE=$RELEASE` like `gcloud run deploy` does with Cloud Run
+
+## Local Run with AI Suggestions
+
+1. Add your OPENAI_API_KEY= to flask .env (personal one ATM)
+2. Start flask locally (./deploy.sh --env=local flask)
+3. Take note of where it's running (likey http://127.0.0.1:8080)
+4. Change NEXT_PUBLIC_FLASK_BACKEND to the url from Step 3
+5. Run the next server (npm run dev)
```
realkosty, line 185: `./deploy.sh react flask` will do these steps automatically for you

flask/src/main.py
```diff
@@ -51,7 +53,8 @@ def before_send(event, hint):
 
 def traces_sampler(sampling_context):
     sentry_sdk.set_context("sampling_context", sampling_context)
-    REQUEST_METHOD = sampling_context['wsgi_environ']['REQUEST_METHOD']
+    wsgi_environ = sampling_context.get('wsgi_environ', {})
+    REQUEST_METHOD = wsgi_environ.get('REQUEST_METHOD', 'GET')
```
realkosty, line 57: is there a legit scenario where wsgi_environ or REQUEST_METHOD are null?

next/src/app/api/suggestion/route.js
```diff
@@ -0,0 +1,40 @@
...
+import { NextResponse } from "next/server";
+import { getProductsOnly } from '@/lib/data.js';
+import {
+  determineBackendUrl,
+} from '@/src/utils/backendrouter';
+
+
+export async function GET(request) {
+  console.log("Sending ai suggeestion request...")
+  const geo = request.nextUrl.searchParams.get("geo");
```
realkosty, line 11: we're implementing this in Next only, not React?

README.md
lukemun: Nice! Looks like running `./deploy.sh flask next` gives me this backend url which fails

```
Supported backend types:  {
  flask: 'http://localhost:8080',
```

but running them separately correctly assigns it and it succeeds.

flask/src/main.py
realkosty: super weird, whatever.

README.md
realkosty: @lukemun this is likely because NEXT_LOCAL_PORT is missing from your `env-config/local.env`

please also update these two:
* https://github.com/sentry-demos/empower/blob/master/env-config/example.env 
* https://github.com/sentry-demos/empower-config/blob/main/local.env (separate repo)

lukemun: Adding to example.env, will put up another CR for local.env

flask/src/main.py
```diff
@@ -96,6 +99,39 @@ def __init__(self, import_name, *args, **kwargs):
 app = MyFlask(__name__)
 CORS(app)
 
+client = OpenAI(api_key= os.environ["OPENAI_API_KEY"])
```
realkosty, line 102: you need to use `os.getenv()`
`os.environ[]` raises KeyError when variable not set
as is it would break production (and local when API key not set)

realkosty (changes requested):

flask/src/main.py
realkosty: and the other spot too
and i think this needs to be moved into a route so we don't initialize OpenAI with None for every request that's not /suggestion

README.md
realkosty: Upon another look I don't think this is necessary, sorry I got confused earlier. See my slack message

flask/src/main.py
lukemun: the client just get's initialized once right? not on every route request

realkosty: probably... not sure how gunicorn manages its objects
but regardless of whether it's once or not it will fail to initialize if key is None (which we decided will be the case in non-local env)

realkosty: if you want it to be more efficient you can put it under a conditional that checks if key exists instead of putting it inside the route

lukemun: Putting it in the route, simpler.

README.md
lukemun: Add the param from slack and got it working

realkosty (approved):

PR #635, opened 2024-12-04
Test PR for bundle analysis

hazal-karakus:


deploy.sh
```diff
@@ -185,6 +185,7 @@ for proj in $projects; do # bash only
   # but that will not work when deploying to staging or production.
   if command -v gcloud &> /dev/null ; then
     export SENTRY_AUTH_TOKEN=$(gcloud secrets versions access latest --secret="SENTRY_AUTH_TOKEN")
+    #export CODECOV_TOKEN2=$(gcloud secrets versions access latest --secret="CODECOV_TOKEN")
```
realkosty, line 188: iirc this is only needed for ASP.NET, where debug file upload happens inside GCP instead of GitHub Actions.

react/config-overrides.js.template
```diff
@@ -21,5 +22,12 @@ module.exports = function override(config, env) {
       reactComponentAnnotation: {enabled:true},
     })
   );
+  config.plugins.push(
+    codecovWebpackPlugin({
+      enableBundleAnalysis: process.env.CODECOV_TOKEN !== undefined,
```
realkosty, line 27: nice

react/src/components/Products.jsx
```diff
@@ -7,6 +7,7 @@ import measureRequestDuration from '../utils/measureRequestDuration';
 import Loader from 'react-loader-spinner';
 import ProductCard from './ProductCard';
 import { useState, useEffect } from 'react';
+import { LuWheatOff, LuSalad } from "react-icons/lu";
```
realkosty, line 10: same

package.json
```diff
@@ -1,5 +1,9 @@
 {
   "dependencies": {
+    "react-icons": "^5.4.0",
```
realkosty, line 3: is this necessary?

realkosty (approved):
Great work. Feel free to ship once minor comments resolved

react/src/components/Products.jsx
hazal-karakus: I tried this to see if the bundle size would increase

realkosty: ah I see. So this PR itself will be used in demo?
Shouldn't we then expect some kind of codecov bot comment here about bundle size?

realkosty (changes requested):
Looks like codecov report generation is failing
that's why we're not getting any codecov bot comments in the PR
I don't see anything in your change that would cause that...
but previous change to `react` was working fine 

https://github.com/sentry-demos/empower/actions/runs/12162257295/job/33918757830?pr=635

realkosty:
@hazal-karakus please try adding these 

--debug
--useStderr
--verbose  

to https://github.com/sentry-demos/empower/blob/88f919fa17e75a5d16a3feaefe1b347c20a868f5/.github/workflows/run_codecov_on_pull_request.yml#L23

to understand why it's returning with non-zero code

react/src/components/Products.jsx
hazal-karakus: I'm testing this locally as well:
 npm test -- --coverage --reporters=jest-junit 

Thanks for tips about debugging, I'll try those as well

PR #649, opened 2024-12-30
Added bundle analysis to the config template of react project

hazal-karakus:
### Done
------
- Added @codecov/webpack-plugin for Bundle Analysis.
- Added build step to the codecove.yml file for BA.
image

### Next Step
---
- Raise another PR to see BA in a PR comment


react/package.json
```diff
@@ -55,6 +55,8 @@
   "devDependencies": {
     "@acemarke/react-prod-sourcemaps": "^0.2.1",
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
+    "@codecov/bundle-analyzer": "^1.6.0",
```
nicholas-codecov, line 58: You're safe to remove this package, as you're using a bundler you only need that package for BA to work.

hazal-karakus: removed it, thanks!

.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -16,10 +16,13 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: "18"
-
+          
       - name: Run React tests
+        env:
+            CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: |
           npm install
```
realkosty, line 24: can remove this line now because build.sh already does that (npm ci)

react/build.sh
```diff
@@ -10,5 +10,5 @@ rm -rf build
 # npm ci does not update minor versions ->
 # (1) less chance of breaking (2) less noise in PR from package-lock.json
 npm ci
-npm run build # defined in 'scripts' in package.json
+CI=false npm run build # defined in 'scripts' in package.json
```
realkosty, line 13: nice!
Please remove `unset CI` from deploy.sh which is now redundant + add comment why CI=false
your way is actually better too because it doesn't permanently unset it.

react/config-overrides.js.template
```diff
@@ -18,8 +18,14 @@ module.exports = function override(config, env) {
       ignoreFile: '.sentrycliignore',
       ignore: ['webpack.config.js'],
       configFile: 'sentry.properties',
-      reactComponentAnnotation: {enabled:true},
+    })
+  );
+  config.plugins.push(
+    codecovWebpackPlugin({
+      enableBundleAnalysis: process.env.CODECOV_TOKEN !== undefined,
```
realkosty, line 25: ideally we'd have some other condition on the environment here (e.g. `process.env.CI` `process.env.REACT_APP_ENVIRONMENT`) otherwise this can mask configuration issues (CODECOV_TOKEN not set by mistake)

maybe something like this so you can test it locally if you want to by setting the token?
```
process.env.CI !== undefined || process.env.CODECOV_TOKEN !== undefined
```

realkosty (changes requested):
🔥

.github/workflows/run_codecov_on_pull_request.yml
hazal-karakus: removed it thanks!

react/build.sh
hazal-karakus: But it would require me to add CI=false to other projects' build.sh files as well right? Because deploy.sh does it for multiple projects

react/config-overrides.js.template
hazal-karakus: Shouldn't it be "&&" instead of "||"? Because otherwise we would enable BA if CI has a value regardless of the CODECOV_TOKEN.
`process.env.CI !== undefined && process.env.CODECOV_TOKEN !== undefined`

realkosty: > otherwise we would enable BA if CI has a value regardless of the CODECOV_TOKEN.
yes that way it won't silently skip the step leaving us scratching our heads and debugging why it's not working.
it should fail in CI if CODECOV_TOKEN is not set, ideally in a way that's obvious how to fix it.

react/build.sh
realkosty: Ah, good point. There's a non-zero chance of build tools for other languages/frameworks also look at this variable...
Could you please then move it to *.yml then where we call it?
`CI=false ./build.sh`

realkosty (approved):
Nice work figuring this all out :)

PR #661, opened 2025-01-10
Changed Codecov settings to upload analysis to master branch

hazal-karakus:
Changed Codecov settings to upload analysis to master branch

I didn't exclude other projects' tests as well for pushing to master branch.

### DIFF
--- a/.github/workflows/run_codecov_on_pull_request.yml
+++ b/.github/workflows/run_codecov_on_pull_request.yml
@@ -1,23 +1,30 @@
 name: codecov.yml
 # Run Codecov on PR open or change
-on: pull_request
+on: 
+  push:
+    branches:
+      - main
+  pull_request:
+
 jobs:
   codecov:
     runs-on: ubuntu-20.04
 
     steps:
       - name: Check out this repository code
+        if: github.event_name == 'pull_request'
         uses: actions/checkout@v3
         with:
           path: empower
           fetch-depth: 0
 
       - name: Install Node
+        if: github.event_name == 'pull_request'
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -1,23 +1,30 @@
...
     steps:
       - name: Check out this repository code
+        if: github.event_name == 'pull_request'
         uses: actions/checkout@v3
         with:
           path: empower
           fetch-depth: 0
 
       - name: Install Node
+        if: github.event_name == 'pull_request'
```
nicholas-codecov, line 22: You'll still probably need these steps before running BA stuff so the action is properly setup.

.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -1,23 +1,30 @@
 name: codecov.yml
 # Run Codecov on PR open or change
-on: pull_request
+on: 
+  push:
+    branches:
+      - main
```
nicholas-codecov, line 6: Looking at where this PR is pointing, i'm assuming you folks have the default branch for this repo set to `master` so this line should be updated to represent that 👀

hazal-karakus: 🤦‍♀️ forgot it, thanks!

.github/workflows/run_codecov_on_pull_request.yml
hazal-karakus: you're right, thank you!

cstavitsky (approved):
Haven't tested it, but looks good to me!

PR #666, opened 2025-01-22
[flask, next] Add Celery Queue for subscribe send email workflow

lukemun:
### Changes
- Setups a celery queue backed by redis
- Updates the subscribe feature in nextjs/react to do the following depending on error_boundary=true/false
    - Send a enqueue request to flask
    - Showcase error boundary
- Adds enqueue endpoint to flask
- Update run.sh to start celery with the flask server
- Added test_subscribe tda test

Subscribe task has a 25% chance of failure. Failures will be retried so published = processed. 

image

### Testing 
- Start flask server and next server (./deploy.sh --env=local ...)
- Got to home page
- Subscribe at footer, an event check queue endpoint is called and task run
- set error_boundary=true, check existing functionality runs
- Ran tda in staging
Screenshot 2025-03-03 at 4 07 32 PM

Tested in staging successfully.

### DIFF
--- a/flask/run.sh
+++ b/flask/run.sh
@@ -12,27 +12,25 @@ source venv/bin/activate
 pip3 install -r requirements.txt
 
 # Store PIDs for cleanup
-REDIS_RELAY_PID=""
 CELERY_PID=""
-FLASK_PID=""
+
 
 function cleanup {
-  # First, kill the specific processes we started
+
+  # Kill Celery worker by PID since it doesn't listen on a port
+  # to be detected by stop.sh
   if [ -n "$CELERY_PID" ]; then
     kill -9 $CELERY_PID 2>/dev/null || true
   fi
 
-  if [ -n "$REDIS_RELAY_PID" ]; then
-    kill -9 $REDIS_RELAY_PID 2>/dev/null || true
-  fi
+  # Kill SSH tunnel by port
 
-  if [ -n "$FLASK_PID" ]; then
-    kill -9 $FLASK_PID 2>/dev/null || true
-  fi
+  stop.sh google_compute_engine $FLASK_LOCAL_REDISPORT
+  stop.sh python3 $LOCAL_PORT
 }
 
 # Register cleanup for all possible termination signals
-trap cleanup EXIT INT TERM HUP QUIT
+trap cleanup EXIT
 
 # needed for caches since GCP redis doesn't have public IP
 ACTIVE_ACCOUNT=$(gcloud auth list --format="value(account)" --filter="status:ACTIVE")
@@ -51,8 +49,7 @@ fi
 # Set up SSH tunnel to the cloud Redis instance
 echo "Setting up SSH tunnel to Redis server at $FLASK_REDIS_SERVER_IP:6379"
 gcloud compute ssh redis-relay --zone=us-central1-a -- -N -L $FLASK_LOCAL_REDISPORT:$FLASK_REDIS_SERVER_IP:6379 &
-REDIS_RELAY_PID=$!
-echo "SSH tunnel started with PID: $REDIS_RELAY_PID"
+
 
 # Give the SSH tunnel time to establish
 sleep 3
@@ -65,9 +62,5 @@ echo "Celery worker started with PID: $CELERY_PID"
 
 # Run Flask in the background and capture its PID
 echo "Starting Flask server on port $LOCAL_PORT"
-flask run --port $LOCAL_PORT &
-FLASK_PID=$!
-echo "Flask server started with PID: $FLASK_PID"
+flask run --port $LOCAL_PORT
 
-# Wait for Flask to exit
-wait $FLASK_PID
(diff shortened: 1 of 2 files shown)


next/src/app/page.jsx
```diff
@@ -21,7 +21,9 @@ export default function Page(props) {
 
   console.log("in home page");
   const router = useRouter();
-  const { backend, frontendSlowdown } = useSearchParams();
+  const searchParams = useSearchParams();
```
lukemun, line 24: frontendSlowdown wasn't used anywhere

realkosty (commented):
Just got some additional direction on this:
we want to add the same to React (could be separate PR) and modify an existing TDA test that already loads home page to add some subscribe clicks.

lukemun:
> Just got some additional direction on this: we want to add the same to React (could be separate PR) and modify an existing TDA test that already loads home page to add some subscribe clicks.

Ok -- I'll break that out in a separate CR

realkosty (commented):
is it expected that Published > Processed? not sure I understand why

generally please include a Sentry UI screenshot in the PR to show what's expected since events are eventually deleted and it's then hard to recover how a particular feature demo is supposed to look in the UI


Screenshot 2025-02-21 at 9 57 02 AM

flask/run.sh
```diff
@@ -11,22 +11,68 @@ fi
 source venv/bin/activate
 pip3 install -r requirements.txt
 
+# Store PIDs for cleanup
+REDIS_RELAY_PID=""
+CELERY_PID=""
+FLASK_PID=""
+
 function cleanup {
```
lukemun, line 19: Whenever I terminated flask locally with CTRL+C, the service (and worker) still ran. Not sure why flask was still running but added this to ensure it got cleaned up.

realkosty: i think there was a bug in stop.sh
i fixed it here https://github.com/sentry-demos/empower/pull/697
did you rebase?

lukemun: I rebased and ran again. Still encountered the issue. 

```
> ENVIRONMENT test
 * Debug mode: off
Address already in use
Port 8080 is in use by another program. Either identify and stop that program, or start the server with a different port.
Terminated: 15
./deploy.sh [ERROR]: flask/run.sh exited early, must be a crash.
```

Interesting. 1st run success. 2nd failure. 3rd success. So the teardown/cleanup of the 2nd seems like it's killing it. 

Fine with reverting this file for now to deal with this as a follow up.

realkosty: I will debug tomorrow

flask/src/queues/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+def sendEmail(self, email, headers=None):
+  try:
+    x = random.randrange(5)
+    if x == 0:
+        raise Exception("sending email error")
+    time.sleep(x)
+    print("Sending email to: " + email)
+    return x
+  except Exception as e:
+      with sentry_sdk.start_transaction(name="src.queues.retry"):
```
realkosty, line 16: nice!
something like this would be even more clear:

`name="src.queues.tasks.sendEmail (retry)"`

probably something like `self.name` or `f"{inspect.getmodule(self).__name__}.{self.__qualname__}"`

flask/run.sh
lukemun: yes 8080 is my FLASK_LOCAL_PORT

flask/src/queues/tasks.py
lukemun: updated.

flask/run.sh
realkosty: @lukemun
okay, I tested it and seeing the same alternating behavior as you described

but also seeing something else (happens every time)

Screenshot 2025-03-11 at 7 23 22 PM

so
do we need to do `REDISPORT=$FLASK_LOCAL_REDISPORT celery -A ...` ???

i'm also confused about the first error `has no attribute email_subscribe`

let's fix these first maybe they are causing a kind of exit that doesn't generate either of the signals you are trapping?

flask/run.sh
```diff
@@ -29,4 +36,8 @@ fi
 
 gcloud compute ssh redis-relay --zone=us-central1-a -- -N -L $FLASK_LOCAL_REDISPORT:$FLASK_REDISHOST:6379 &
 
+sleep 1 
+
+celery -A src.queues.email_subscribe worker -l INFO &
```
realkosty, line 41: Also, don't wee need to kill this one too?

IIRC when you use & to background a process it continue running even if the parent shell script terminates.

flask/app.yaml.template
```diff
@@ -1,11 +1,12 @@
 service: ${SERVICE}
 runtime: python312
-entrypoint: gunicorn -b :\$PORT -w 2 --timeout 60 src.main:app
+entrypoint: ./start.sh
+
+instance_class: F2
```
realkosty, line 5: $37 -> $73 /month for the record. Not an issue
Did it help with something?

realkosty (approved):

flask/app.yaml.template
lukemun: was hitting memory / cpu limits with celery running on the instance causing it to be terminated, bumping this resolved it (just exceeding the limit for the smaller class).

PR #667, opened 2025-01-22
[react] feat/restore critical exp

gid-sentry:
Pulling in Sergio's changes with (very slight) modifications in order to be able to restore the critical experience dashboard and build some workflows utilizing the new EAP functionality.


ndmanvar:
@gid-sentry - looks like we have some conflicts that we need to resolve..
and also, let's remove the log statements if we don't need them 😉

gid-sentry:
Conflicts should be fixed now I believe

react/config-overrides.js
```diff
@@ -0,0 +1,25 @@
+const SentryWebpackPlugin = require('@sentry/webpack-plugin');
```
realkosty, line 1: could you please remove `react/config-overrides.js` from commit and add it to .gitignore? (this is an artifact of another new change)

realkosty (changes requested):
Hey I checked it out locally and it's failing to build

Screenshot 2025-01-24 at 3 33 52 PM

gid-sentry:
@realkosty should be all set now!

realkosty:
@ndmanvar

so `npm test --coverage` will return exit code 1 if one of the tests fails.
I think we should fix this once and for all:

https://github.com/sentry-demos/empower/pull/670 (make sure coverage report succeeds even if some test fails)

https://github.com/sentry-demos/empower/pull/669 (actually fixes test FWIW)

realkosty:
> @realkosty should be all set now!

sweet, all looks 👍👍

realkosty (approved):

ndmanvar:
dude we got a supect issue, which confirms this is deployed


bammmm

PR #669, opened 2025-01-29
[react] fix Products.test.js failing after PR 667; rename codecov.yml

realkosty:
Screenshot 2025-01-29 at 2 03 29 PM

although I think we should make `npm test --coverage` pass regardless if tests succeed or fail


realkosty:
self-approving

PR #675, opened 2025-01-30
AWS - SNS publisher and lambda code

lukemun:
- Script for upload 20 sns inventory update messages
  - contains a negative inv value which will throw an error
- lambda function and corresponding libraries which are manually zipped and uploaded to AWS


aws/inventory_update_sns.py
```diff
@@ -0,0 +1,408 @@
...
+      'plantNames': ["fern", "spider", "cactus"],
+      'inventory': [-10, 10, 20],
+      # Add i to keep shipment IDs sequential
+      'shipment_id': 666
+    }
+    messages.append(message)
+
+    return messages
+
+def usage_demo():
```
lukemun, line 366: Start of upload script

aws/lambda_demo/lambda_function.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Current time in UTC
+    now = datetime.now(timezone.utc)
+
+    # Calculate latency
+    latency = now - message_time
+    return latency
+
+@serverless_function
+def lambda_handler(event, context):
```
lukemun, line 41: lambda function (used for both SQS right now)

lukemun (commented):
Added comments to find relevant code easier

.gitignore
```diff
@@ -18,6 +18,14 @@ express/node_modules
 react/build
 next/dist
 
+# Environments
+.env
+.venv
+env/
+venv/
```
realkosty, line 25: 1. these two already exist:

.env
venv 

2. also I think `env/` covers already existing `tda/env`

aws/inventory_update_sns.py
```diff
@@ -0,0 +1,438 @@
...
+import sentry_sdk
+from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
+import boto3
+from botocore.exceptions import ClientError
+from contextlib import contextmanager
+
+dotenv.load_dotenv()
+
+sentry_sdk.init(
+    dsn=os.environ["FLASK_APP_DSN"],
```
realkosty, line 27: why use flask DSN here instead of PYTHON_LAMBDA_SENTRY_DSN?

aws/lambda_demo/lambda_function.py
```diff
@@ -0,0 +1,123 @@
+import json, os
+import sentry_sdk
+from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
+from sentry_sdk.integrations.serverless import serverless_function
+from datetime import datetime, timezone
+from contextlib import contextmanager
+
+
+sentry_sdk.init(
+    dsn=os.environ.get("PYTHON_LAMBDA_SENTRY_DSN"),
```
realkosty, line 10: please rename directory `aws` -> `aws-lambda-python` or `python-aws-lambda` to be more clear

.gitignore
```diff
@@ -91,6 +99,14 @@ env-config/*
...
 tda/env
 
+
+# Lambda demo specific
+aws/lambda_demo/certifi/
+aws/lambda_demo/sentry_sdk/
+aws/lambda_demo/urllib3/
+aws/lambda_demo/*.dist-info/
+aws/lambda_demo/*.egg-info/
+
```
realkosty, line 109: should we also include `aws/lambda_demo/lambda_demo.zip`?

realkosty (changes requested):
Wow 👏👏👏

My main feedback is that it would hard for another person to figure out how to make use of this. Can you please add README.md?

Ideally we'd want to cover:

1. How to deploy this (hopefully we can automate in the future, see `crons-python` for example of non App Engine deploy)
2. How to generate demo data using this code (steps)
3. How to demo using demo data generated with this (ideally including some screenshots)
4. Whether this in any way interacts with other Empower applications or DB (IIUC it does not, is completely isolated in AWS just follows the ecommerce theme)

+ some minor feedback inline

PR #681, opened 2025-02-10
Fix server side instrumentation for Next

landenai:
### Changes

- Replaced prisma with postgres node for N + 1 products query
- Moved reroute logic to clean up UX and more closely mirror realistic ecommerce flow 
- Persisted `se` tag in server runtime

### Steps to Reproduce
- Load product page and get N + 1 error
- Add `se` query param and see tag in server action error

### DIFF
--- a/next/lib/data.js
+++ b/next/lib/data.js
@@ -7,8 +7,6 @@ import { redirect } from 'next/navigation';
 import { query } from '@/lib/db';
 import { cookies } from 'next/headers';
 
-const cookiesStore = await cookies();
-
 const prisma = new PrismaClient();
 
 export async function getProductsRaw() {
@@ -72,6 +70,7 @@ export async function getProduct(index) {
 
 export async function checkoutAction(cart) {
   // Need to set the se tag on the server runtime scope
+  const cookiesStore = await cookies();
   const se = cookiesStore.get("se");
   if(se) {
     Sentry.getCurrentScope().setTag("se", se.value)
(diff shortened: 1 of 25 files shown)


next/lib/data.js
```diff
@@ -3,24 +3,61 @@ import { PrismaClient } from '@prisma/client';
...
 export default async function getProducts() {
   try {
     console.log("Fetching products...");
     // Artificial slowdown for demoing
-    await new Promise((resolve) => setTimeout(resolve, 2000));
-
-    const products = await prisma.products.findMany();
+    //await new Promise((resolve) => setTimeout(resolve, 2000));
+    const sleepDuration = 2;
+    const products = await prisma.$queryRawUnsafe(`
```
lukemun, line 45: does this $queryRawUnsafe generate an N+1 query or do we have not use prisma?

lukemun (approved):
Just the 1 q, otherwise lgtm

next/lib/data.js
landenai: Removed this entire function, replaced by `getProductsRaw()`

lukemun:
Deployment is failing FYI: https://vercel.com/sentry/empower/8BPLBnveEakRC5czLbwLYXVMu2Fk?filter=errors

next/lib/data.js
```diff
@@ -102,7 +69,12 @@ export async function getProduct(index) {
 }
 
 export async function checkoutAction(cart) {
-  
+  // Need to set the se tag on the server runtime scope
```
lukemun, line 72: Just curious -- was this the recommended way to persist the tag to the server side? Lot of customers will run into this

lukemun (approved):

PR #692, opened 2025-02-19
Feature: flaky tests

vlad-ko:
trying to open a PR with flaky tests


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "@sentry/react": "^9.1.0",
```
realkosty, line 3: do you remember the reason for adding this?
pretty sure we don't need package.json in root directory at all...

PR #696, opened 2025-02-21
fix: update the codecov workflow

thomasrockhu-codecov:
fixes https://github.com/codecov/engineering-team/issues/3367

### DIFF
--- a/.github/workflows/run_codecov_on_pull_request.yml
+++ b/.github/workflows/run_codecov_on_pull_request.yml
@@ -12,8 +12,9 @@ jobs:
 
     steps:
       - name: Check out this repository code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
+          path: empower
           fetch-depth: 0
 
       - name: Install Node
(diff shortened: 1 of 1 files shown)


realkosty (approved):

.github/workflows/run_codecov_on_pull_request.yml
```diff
@@ -23,29 +22,29 @@ jobs:
...
+          verbose: true
 
       - name: Run Python Tests
         run: |
           pip install -U pytest
           pip install pytz
           pip install pytest-codecov
           pytest --cov=. --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
-        working-directory: ./empower/flask/src
+        working-directory: ./flask/src
```
realkosty, line 41: @thomasrockhu-codecov just curious, is this somehow necessary for it to work (e.g. ignores `working-directory` when looking for codecov.yml and always looks in cwd) or just aesthetic?

thomasrockhu-codecov: this is necessary because the `checkout` step doesn't shove the repository into `empower`. This has nothing to do with the yml

thomasrockhu-codecov: i.e. if `working-direcotry` is not changed here, the commands above will fail because the directory structure has changed with the removal of 

`path: empower`

realkosty: @thomasrockhu-codecov yes, I saw that :-) 
was the removal of `path: empower` necessary or aesthetic?

thomasrockhu-codecov: @realkosty it is probably possible to get it work with `path: empower`, but I don't see why it was necessary in the workflow to begin with.

PR #699, opened 2025-02-26
Adding subscribe queueing to react

lukemun:
- Updates the subcribe to put an email on the queue via flask
- Adds error_boundary url param
- Adds a test to hit subscribe
- Add manual instrumentation to better show enqueue and retry transactions


flask/run.sh
```diff
@@ -11,22 +11,65 @@ fi
...
 
+# Store PIDs for cleanup
+REDIS_RELAY_PID=""
+CELERY_PID=""
+FLASK_PID=""
+
 function cleanup {
-  stop.sh python3 $LOCAL_PORT
-  stop.sh gcloud 6379
+  echo "Cleaning up processes..."
```
lukemun, line 20: For some reason this cleanup wasn't happening. Everything I killed the flask process local with CTRL+C and restarted it would fail saying the port was in use. Adding this fixed it (it's AI generated)

PR #704, opened 2025-02-28
Update java sdk to v8.38.0 in spring-boot demo

ndmanvar:


realkosty (approved):
LGTM (tested locally)

although we won't be able to deploy as GCP dropped support for Java 8 apparently

```
[INFO] GCLOUD: ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: Error(s) encountered validating runtime. Your runtime version for java is past End of Support. Please upgrade to the latest runtime version available..
```

but that has nothing to do with this change and needs to be fixed separately

PR #733, opened 2025-03-15
upgrade to java 17 and spring 2.7.18

ndmanvar:


ndmanvar:
tested /products locally
image

PR #735, opened 2025-03-16
detect blocking calls on aspnet core

bruno-garcia:
More context:

* https://www.jamescrosswell.dev/posts/async-detecting-blocking-calls/
* https://github.com/getsentry/sentry-dotnet/pull/2709
* https://x.com/brungarc/status/1711484812702171309

If suspect commit is set up, should point to my commit? 🤞

### DIFF
--- a/aspnetcore/Controllers/ProductsController.cs
+++ b/aspnetcore/Controllers/ProductsController.cs
@@ -21,7 +21,7 @@ public class ProductsController : ControllerBase
     {
         var products = await GetProductsAsync();
         
-        await CallHttpApiAsync();
+        CallHttpApiAsync().Wait();
         
         return products;
     }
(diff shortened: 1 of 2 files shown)


aspnetcore/Controllers/ProductsController.cs
```diff
@@ -55,12 +55,12 @@ private async Task CallHttpApiAsync()
         try
         {
             response.EnsureSuccessStatusCode();
+            span?.Finish();
         }
         catch (Exception exception)
         {
+            span?.Finish(exception);
```
bruno-garcia, line 62: Unrelated to the blocking feature. But we should be finishing the span in the `catch` block and there's a convenient overload that takes exceptions for this reason.

realkosty: nice! didn't know that

realkosty (approved):

PR #738, opened 2025-03-18
[platform] fix new envsubst.sh that was breaking on ubuntu-22.04

realkosty:
fixes https://github.com/sentry-demos/empower/actions/runs/13914950356/job/38936202103

# Testing
`./deploy.sh --env=staging flask`


realkosty:
merging to test it works on ubuntu

PR #741, opened 2025-03-18
[tda] Removes query param from the url 

lukemun:
- Subscribe tda test was cycling through backend query params. 
- However, the only backend with the subscribe queue feature is flask. 

Testing
- Deployed to sauce locally and saw successful test run

### DIFF
--- a/tda/desktop_web/test_subscribe.py
+++ b/tda/desktop_web/test_subscribe.py
@@ -6,7 +6,7 @@ from selenium.webdriver.common.by import By
 from selenium.common.exceptions import NoSuchElementException
 
 
-def test_subscribe(desktop_web_driver, endpoints, batch_size, backend, sleep_length, random=None):
+def test_subscribe(desktop_web_driver, endpoints, batch_size, sleep_length, random=None):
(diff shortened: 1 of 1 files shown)


tda/desktop_web/test_subscribe.py
```diff
@@ -16,26 +16,9 @@ def test_subscribe(desktop_web_driver, endpoints, batch_size, backend, sleep_len
         sentry_sdk.set_tag("endpoint", endpoint)
 
         for i in range(batch_size):
-
-            # Build a query string to pass the chosen backend, if your app needs it
-            # for the subscription endpoint. Adjust as needed.
-            query_string = {
-                'backend': backend()
```
realkosty, line 23: maybe set explicitly to flask for clarity and if default changes at some point?

realkosty (approved):

realkosty (approved):

tda/desktop_web/test_subscribe.py
lukemun: updated

PR #743, opened 2025-03-20
Add statsig feature flag to react + flask

lukemun:
## Changes
- Add Statsig secret fetching to deploy script
- Update the sdks to latest so feature flags are available
- Initialize in root and checks all flags
- Each flag value is configurable in statsig UI

Tested locally with checkout flow

### **Before**
image

### **After**
image

image

### **Before**
image

### **After**
image

image


package.json
```diff
@@ -1,5 +1,7 @@
 {
   "dependencies": {
+    "@sentry/browser": "^9.7.0",
+    "@sentry/react": "^9.7.0",
```
realkosty, line 4: can you please remove ./package.json and ./package-lock.json

react/package.json
```diff
@@ -3,7 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@sentry/react": "~8.32.0",
+    "@sentry/react": "^9.7.0",
```
realkosty, line 6: that's a big upgrade, we might want to go through https://docs.sentry.io/platforms/javascript/migration/v8-to-v9/ item-by-item or at least inspect demo data for any regressions.

react/src/index.js
```diff
@@ -68,10 +69,19 @@ let CHECKOUT_SUCCESS;
 let ERROR_BOUNDARY;
 const DSN = process.env.REACT_APP_DSN;
 const RELEASE = process.env.REACT_APP_RELEASE;
-
+const STATSIG_CLIENT_KEY = process.env.STATSIG_CLIENT_KEY;
```
realkosty, line 72: please add to `react/validate_env.list`

realkosty (changes requested):
nice

react/package.json
lukemun: How's this? :p also tested locally and didn't run into issues yet, will keep an eye on sentry projs
image

realkosty: Nice idea but not super impressed with Cursor's performance here :)
it only checked like 5 things out of ~ 20-30
i searched manually for a bunch more and don't see matches

please include links to or even better screenshots of flagship error + flagship transaction, BEFORE and AFTER each

realkosty: Those should be included in the description of every PR that makes major changes, SDK or platform upgrades

lukemun: Ok, added it. It's bit hard to show with screenshots and the issue detail page has gotten so big but it's there now.

deploy.sh
```diff
@@ -118,6 +118,48 @@ for proj in $projects; do # bash only
   echo "||| $0: $proj"
   echo "|||"
 
+  # Function to fetch secrets from Google Cloud Secret Manager (uses same name for env var and secret)
+  fetch_secrets_to_env() {
+    # Check if gcloud is available
+    if ! command -v gcloud &> /dev/null; then
+      echo "[WARN] gcloud command not found, skipping secret fetching"
```
realkosty, line 125:
```suggestion
      echo "[ERROR] gcloud command not found - required to fetch secrets"
```

deploy.sh
```diff
@@ -118,6 +118,48 @@ for proj in $projects; do # bash only
...
+      else
+        # Export the secret to environment
+        export "$secret_name=$value"
+        echo "  Successfully set $secret_name"
+      fi
+    done
+  }
+
+  # Fetch secrets before validating project
+  fetch_secrets_to_env
```
realkosty, line 161: do we need to do this when --env !=`local` ?

deploy.sh
```diff
@@ -118,6 +118,48 @@ for proj in $projects; do # bash only
...
+    echo "Fetching secrets from Google Cloud Secret Manager..."
+
+    # Define the list of secrets to fetch - names will be identical in env and Secret Manager
+    # TODO: Incrementally add to this and remove from .env and validate_env.list
+    local secrets=(
+      # Sentry CLI will use this token for authentication.  Otherwise, one can use `sentry-cli login`,
+      # but that will not work when deploying to staging or production.
+      "SENTRY_AUTH_TOKEN"
+      "STATSIG_CLIENT_KEY"
+      "STATSIG_SERVER_KEY"
```
realkosty, line 138: this is fine for now, but in the future i'd like to move this list to `*.env` with some kind of convention indicating it's fetched from secret manager
`SENTRY_AUTH_TOKEN=****`

deploy.sh
```diff
@@ -118,6 +118,48 @@ for proj in $projects; do # bash only
...
+    # Check if gcloud is available
+    if ! command -v gcloud &> /dev/null; then
+      echo "[WARN] gcloud command not found, skipping secret fetching"
+      return 1
+    fi
+
+    echo "Fetching secrets from Google Cloud Secret Manager..."
+
+    # Define the list of secrets to fetch - names will be identical in env and Secret Manager
+    # TODO: Incrementally add to this and remove from .env and validate_env.list
```
realkosty, line 132: we should keep them in validate_env.list
they way it's implemented now does validate_env check happen after you get the secrets?

deploy.sh
realkosty: previous SENTRY_AUTH_TOKEN fetch code (line 183) was for `aspnetcore` which has a bit of a weird setup - so maybe it is needed for that one at least in staging/production

realkosty (changes requested):

realkosty (changes requested):
Also I think we're missing something like this in the yaml template:

```
env_variables:
  MY_API_KEY: secure:my-secret-key
  DATABASE_PASSWORD: secure:projects/YOUR_PROJECT_ID/secrets/database-password/versions/latest
  ```

please both of --env=local and --env=staging

File: deploy.sh (hunk 7 above)
Comment (lukemun) [reply deploy.sh:132]: yes on generated_envs+="$(../env.sh $env) "

File: deploy.sh (hunk 6 above)
Comment (lukemun) [reply deploy.sh:138]: ok sg

File: deploy.sh (hunk 5 above)
Comment (lukemun) [reply deploy.sh:161]: do we want full functionality locally? feature flag and any redis related features will break when we migrate those secrets

Comment (lukemun) [reply deploy.sh:161]: oh wait you send env != local, then yes that's the whole point

Comment (realkosty) [reply deploy.sh:161]: no I mean, when env **!=** local, i.e. staging/production
I thought the idea was that they will be automatically wired by GCP through some kind of app.yaml magic?

Hm I guess if we do it this way there's less changes needed as it handles all environments, local and GCP.
But that would mean they are passed form GH Actions environment to GCP via `.env` file upload which is not very secure, so we want to avoid that.

Comment (lukemun) [reply deploy.sh:161]: i see, should be doable with app.yaml but may start to be a pain adding all those. 

can we keep this simple, iterative, and have that as a next step? else this may continue to drag and this should already be a security improvement

Comment (realkosty) [reply deploy.sh:161]: > automatically wired by GCP through some kind of app.yaml magic?
did some research, turns out that's not a thing
need wrapper script that called in app.yaml as entrypoint, e.g.

```
entrypoint: entrypoint.sh <regular entrypoint command e.g. gunicorn/node/etc>
```

Comment (realkosty) [reply deploy.sh:161]: > can we keep this simple, iterative, and have that as a next step? else this may continue to drag and this should already be a security improvement

okay, let's roll with this for now

Comment (lukemun) [reply deploy.sh:161]: Could to take a look [at]( https://github.com/sentry-demos/empower/pull/801)? 

I moved the secret logic there, this branches off that and will just have the FF work

**Comment by lukemun** [comment]
@realkosty For a reason I cannot figure out, the react statsig client always returns false for the gate checks. On the statsig UI it says the client wasn't initialized but it is... I checked the secret value, docs, etc.

Everything still works fine so this is not blocking we'll just have false values in the react project for now.  

Posted in the Statsig Slack support, waiting for a response.

**Review by realkosty — CHANGES_REQUESTED** [review]
currently when not authenticated in gcloud this will fail in a way that's unclear why
could you please add something like this?

```
if command -v gcloud &> /dev/null ; then
  ACTIVE_ACCOUNT=$(gcloud auth list --format="value(account)" --filter="status:ACTIVE")
  if [ -z "$ACTIVE_ACCOUNT" ]; then
    echo "You are not authenticated with Google Cloud. Press any key to authenticate..."
    read -n 1 -s
    gcloud auth login
  fi
fi
```

**Comment by lukemun** [comment]
> currently when not authenticated in gcloud this will fail in a way that's unclear why could you please add something like this?
> 
> ```
> if command -v gcloud &> /dev/null ; then
>   ACTIVE_ACCOUNT=$(gcloud auth list --format="value(account)" --filter="status:ACTIVE")
>   if [ -z "$ACTIVE_ACCOUNT" ]; then
>     echo "You are not authenticated with Google Cloud. Press any key to authenticate..."
>     read -n 1 -s
>     gcloud auth login
>   fi
> fi
> ```

Added at the top of deploy (with some more logs), since we shouldn't proceed at if we're just going to error out later

File: deploy.sh [hunk 8]
```diff
@@ -49,6 +49,24 @@
 set -e # exit immediately if any command exits with a non-zero status
 # https://fvue.nl/wiki/Bash:_Error_handling
 
+
+# Check if user is authenticated with Google Cloud before running deployment
+if command -v gcloud &> /dev/null ; then
+  ACTIVE_ACCOUNT=$(gcloud auth list --format="value(account)" --filter="status:ACTIVE")
+  if [ -z "$ACTIVE_ACCOUNT" ]; then
+    echo "You are not authenticated with Google Cloud. Press any key to authenticate..."
```
realkosty, line 57: So i noticed that sometimes even when account is active it will still prompt to re-authenticate. Something like this should cover that scenario. Hopefully it's not too brittle
```suggestion
  if [ "$(gcloud auth print-access-token 2>/dev/null | wc -c)" -le 200 ]; then
    echo "You are not authenticated with Google Cloud. Press any key to authenticate... (browser window may open)"
```

realkosty (changes requested):
see suggestion + 
failing in staging with 
```
DEFAULT 2025-05-06T19:55:47.752101Z statsig.statsig_errors.StatsigValueError: Invalid key provided. You must use a Server Secret Key from the Statsig console
```

deploy.sh
lukemun: fine with that but lets keep more deploy.sh refactors for later please, would like to get this FF out.

lukemun:
> see suggestion + failing in staging with
> 
> ```
> DEFAULT 2025-05-06T19:55:47.752101Z statsig.statsig_errors.StatsigValueError: Invalid key provided. You must use a Server Secret Key from the Statsig console
> ```

I can't replicated this. Deployed to staging myself and am not seeing that error. Looks like it only happened twice.

Updated flask statsig initialization so failures should block anything.

realkosty (approved):

PR #751, opened 2025-03-24
[flask] Fix error rate by removing custom instrumentation

lukemun:
- Custom instrumentation was creating a transaction within a transaction, resulting in the sdk not picking up the real error rate
- Remove custom instrumentation means the sdk will infer the name
-- retry now has same transaction name as consumer
-- main produce is named after function endpoint "enqueue"

**Before**
image

**After**
image

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -124,11 +124,9 @@ redis_client = redis.Redis(host=redis_host, port=redis_port, decode_responses=Tr
 @app.route('/enqueue', methods=['POST'])
 def enqueue():
     body = json.loads(request.data)
-    print(body['email'])
     email = body['email']
-    with sentry_sdk.start_transaction(name="src.api.enqueueEmail"):
-        r = sendEmail.apply_async(args=[email], queue='celery-new-subscriptions')
-        print(r.task_id)
+    r = sendEmail.apply_async(args=[email], queue='celery-new-subscriptions')
+    print(f"task id: {r.task_id} for email: {email}")
     return jsonify({"status": "success"}), 200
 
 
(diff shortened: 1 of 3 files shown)


realkosty (commented):
Oh interesting. I wonder why it works that way...

Don't you think that error rate for src.queues.tasks.sendEmail made sense in Before but doesn't make sense in After (`(1 - 6/6) != 16.67%` or `6 + 1 != 7`) or am i missing something?

should we perhaps instead use something like this:

```
scope = sentry_sdk.get_current_scope()
scope.set_transaction_name("src.api.enqueue")
```
in main.py, and not touch tasks.py at all?

lukemun:
> Oh interesting. I wonder why it works that way...
> 
> Don't you think that error rate for src.queues.tasks.sendEmail made sense in Before but doesn't make sense in After (`(1 - 6/6) != 16.67%` or `6 + 1 != 7`) or am i missing something?
> 
> should we perhaps instead use something like this:
> 
> ```
> scope = sentry_sdk.get_current_scope()
> scope.set_transaction_name("src.api.enqueue")
> ```
> 
> in main.py, and not touch tasks.py at all?

I think it's 1 failure in 6 messages => 1/6 = 16.67%. 

The error rate displayed for `src.queues.tasks.sendEmail` consumer was not impacted by this change. It's 20% in code so just some variance here. 

The issue was in the producers which always had 100% error rate, that is now fixed for main producer (enqueue). Though the `src.queues.tasks.sendEmail` retry producer still has 100% error rate (not sure that's correct). 

Regardless, using the auto instrumentation is what Anton recommended (custom + auto is unsupported) so I think we should go with the auto and report issues to the team later.

realkosty (approved):

PR #753, opened 2025-03-25
Catch Inventory Error for CEXP

serglom21:
Before 
Screenshot 2025-03-25 at 1 56 24 PM

After
Screenshot 2025-03-25 at 1 56 52 PM

The fix to this is quite simple. But I am having some issues getting source maps working to show the working error's stack trace that are not related to the changes I am making here :(

### DIFF
--- a/react/src/components/ProductCard.jsx
+++ b/react/src/components/ProductCard.jsx
@@ -16,7 +16,11 @@ function ProductCard(props) {
   const stars = props.stars;
 
   function validate_inventory(product) {
-    return product && inventory.includes(product.id)
+    try {
+      return product && inventory.includes(product.id)
+    } catch (error) {
+      throw new Error("Error validating inventory");
+    }
   }
 
   return (


realkosty:
Hey @serglom21, do you remember what was the purpose of this change?

would you mind adding more context to the description? (also I'm trying to use it as a "model" PR because it has nice BEFORE/AFTER screenshots)

serglom21:
@realkosty I added a description. Let me know if that is clear / you need more :)

realkosty:
@serglom21 thanks!

> The [CEx issue](https://demo.sentry.io/issues/6713516869/events/1bcaf1c1443248efa6fd8febfce74805/) does not report a useful error message. Its not evident that there is an inventory error without looking at the stack trace.

hm, i don't get it though - it's a programming error, `inventory` variable is not defined in that scope. How is that not useful? People don't always put in try-catch blocks in real life, esp. when there is no "expected" error that can happen (aka Java's checked exceptions) or if there is no good way to recover from the error

serglom21:
@realkosty I agree with that. Realistically, there are a lot of unhandled crashes as the underlying cause of serious issues. I am ok with leaving this error be "realistic" by leaving it how it is or also by making it more actionable - which in cases, devs might also choose to handle an error gracefully. More so, in methods that are critical like `validate_inventory`.

I am ok with whatever path is chosen here :)

PR #756, opened 2025-03-26
ref(cli): Don't use deprecated `sentry-cli` command

szokeasaurusrex:
`sentry-cli releases files upload-sourcemaps` has been deprecated in favor of `sentry-cli sourcemaps upload` for quite some time now. Keeping this code around in the demo repo has confused folks who are unaware of this deprecation.


realkosty (approved):
Thank you!

PR #780, opened 2025-04-11
chore: upgrade Node.js version to 20.x.x in Express service

ndmanvar:
GCP has announced deprecation of node18, therefore we need to upgrade


ndmanvar:
image
https://demo.sentry.io/insights/backend/?project=4509136302440448&statsPeriod=24h

realkosty (approved):

PR #794, opened 2025-04-22
Pretty sure we don't need this variable...

realkosty:
### Very small change, no need to test!


bruno-garcia (approved):
No tests is totally fine. YOLO :shipit:

PR #799, opened 2025-04-25
Updated URLs for Session Replay network details

hazal-karakus:
Added all urls to `networkDetailAllowUrls` configuration to see Response and Request information at Session Replay Network -> Response/Request tabs


realkosty (approved):
Nice! 

next time please run `./deploy.sh --env=staging react` and provide a sample event link in the PR, or even better - screenshot

PR #801, opened 2025-04-28
Fix failing staging deployments + google secret manager

lukemun:
image

- migration to team-se requires new sentry auth token
- added auth token in gcloud
- updated script to conditionally fetch based on env
- moved auth token fetching to be before source maps upload to fix org mismatch deployment bug
  - may have only existed on my local but we need to override whatever is in SENTRY_AUTH_TOKEN with the correct env secret
- updated validate_env for better logging

 TESTING
- Successfully deployed locally + staging

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -187,10 +187,11 @@ for proj in $projects; do # bash only
 
   cd $top/$proj
 
-  # React builds env vars in; Express/Flask need .env at runtime.
+  # React "bakes in" env variables (exported from calling shell) as well as contents 
+  # of .env into the build, whereas Express/Flask need .env at runtime
   # We dynamically generate a temporary .env from env-config/$env.env.
-  # This command calls ../env.sh, which also validates env vars listed in
-  # the project's validate_env.list via bin/validate_dotenv.sh -> bin/validate_env.sh.
+  # env.sh among other things validates env vars listed in the project's validate_env.list
+  # via bin/validate_dotenv.sh -> bin/validate_env.sh.
   generated_envs+="$(../env.sh $env) "
 
   # We do this because 1) we need RELEASE that's generated in env.sh 2) we need *_APP_*_BACKEND
--- a/deploy.sh
+++ b/deploy.sh
@@ -118,18 +118,79 @@ for proj in $projects; do # bash only
   echo "||| $0: $proj"
   echo "|||"
 
+  # Function to fetch secrets from Google Cloud Secret Manager (uses same name for env var and secret)
+  fetch_secrets_to_env() {
+    # Check if gcloud is available
+    if ! command -v gcloud &> /dev/null; then
+      echo "[ERROR] gcloud command not found, skipping secret fetching"
+      return 1
+    fi
+
+    echo "Fetching secrets from Google Cloud Secret Manager for environment: $env..."
+
+    # Define the list of secrets to fetch - names will be identical in env and Secret Manager
+    local secrets=(
+      "STATSIG_CLIENT_KEY"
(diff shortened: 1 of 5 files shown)


deploy.sh
```diff
@@ -118,18 +118,78 @@ for proj in $projects; do # bash only
...
+    for secret_name in "${secrets[@]}"; do
+      echo "  Fetching secret $secret_name..."
+
+      # Fetch the secret value
+      local value
+      value=$(gcloud secrets versions access latest --secret="$secret_name" --project=sales-engineering-sf 2>/dev/null)
+
+      if [ $? -ne 0 ]; then
+        echo "[WARN] Failed to fetch secret '$secret_name', skipping"
+      else
```
realkosty, line 169: Let fail early so it's clear that required secret it missing and we don't have to scratch our heads later why something is not working

```
>&2 echo "[ERROR] Failed to fetch secret '$secret_name'"
exit 1
```

deploy.sh
```diff
@@ -118,18 +118,78 @@ for proj in $projects; do # bash only
...
+    local secrets=(
+      "STATSIG_CLIENT_KEY"
+      "STATSIG_SERVER_KEY"
+    )
+
+    # Add environment-specific Sentry token
+    local sentry_token_name
+    case "$env" in
+      "prod")
+        sentry_token_name="SENTRY_AUTH_TOKEN_PROD"
```
realkosty, line 141: [nit] let's name explicitly SENTRY_AUTH_TOKEN_[DEMO|TEAM_SE]

deploy.sh
```diff
@@ -118,18 +118,78 @@ for proj in $projects; do # bash only
...
-  # See: https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/env.js
-  # and https://create-react-app.dev/docs/adding-custom-environment-variables/
-  #
-  # Express and Flask on the other hand need .env deployed and present at runtime.
-  #
-  # We generate a temporary .env dynamically from env-config/*.env then remove upon exit
+  # React builds env vars in; Express/Flask need .env at runtime.
+  # We dynamically generate a temporary .env from env-config/$env.env.
+  # This command calls ../env.sh, which also validates env vars listed in
+  # the project's validate_env.list via bin/validate_dotenv.sh -> bin/validate_env.sh.
```
realkosty, line 192: is this intentional or cursor decided to reword on its own?

deploy.sh
realkosty, line 141: [nit] let's rename to `SENTRY_AUTH_TOKEN_PROD_DEMO` and `SENTRY_AUTH_TOKEN_STAGING_TEAM_SE` for clarity

realkosty (changes requested):

deploy.sh
lukemun: Intentional, I had to trace through the scripts to find out that env.sh validated the vars so i asked cursor to make this less verbose and add that in the comment.

deploy.sh
```diff
@@ -118,18 +118,79 @@ for proj in $projects; do # bash only
...
```
realkosty, line 193:
```suggestion
  # React "bakes in" env variables (exported from calling shell) as well as contents 
  # of .env into the build, whereas Express/Flask need .env at runtime
  # We dynamically generate a temporary .env from env-config/$env.env.
  # env.sh among other things validates env vars listed in the project's validate_env.list
  # via bin/validate_dotenv.sh -> bin/validate_env.sh.
```

realkosty (approved):
please apply 1 suggestion then can ship

PR #812, opened 2025-05-07
Fix missing depedency for tests

lukemun:


lukemun:
Fixing this [build](https://github.com/sentry-demos/empower/actions/runs/14885501928/job/41803894279)

@realkosty

lukemun:
Can we configure tests to run when a PR is pushed not just on merges?

realkosty:
> Can we configure tests to run when a PR is pushed not just on merges?

Hm I thought we did.. for some reason all the runs are all cancelled
https://github.com/sentry-demos/empower/actions/workflows/run_codecov_on_pull_request.yml

realkosty (approved):

PR #837, opened 2025-05-31
Revert "[tda] increase number of threads 8 -> 16 (#833)"

realkosty:
Something happened when this was deployed..
This reverts commit 407fb85d6067859b8b93c5e4319e493fedc3cebf.

Screenshot 2025-05-30 at 5 30 05 PM
Screenshot 2025-05-30 at 5 29 47 PM

### DIFF
--- a/tda/jobs/desktop_web.sh
+++ b/tda/jobs/desktop_web.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-BATCH_SIZE=random_20 pytest -s -n 13 --ignore-glob='*_vue.py' desktop_web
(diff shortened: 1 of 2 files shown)


realkosty:
when running locally with `-n 16` I see about 0.7-1.1 GB of memory used
this was an e2-micro instance with 1 GB total
also canaries overlap with actual process which can double memory footprint

realkosty:
Had to restart VM because it was completely unresponsive

realkosty:
re-applied in https://github.com/sentry-demos/empower/pull/838

Issue #850, opened 2025-06-06
ASPNET Core: Too many blocking calls reported


bruno-garcia:
https://sandbox.sentry.io/issues/?project=4508160335675392&statsPeriod=14d

Ideally we'd have 1 or 2 blocking examples only. but we have a lot:

![Image](https://github.com/user-attachments/assets/b4045793-e346-494d-84c1-7f17279b0b35)

Some of those are coming from the SDK itself: https://sandbox.sentry.io/issues/6640476210/?project=4508160335675392&query=is%3Aunresolved&referrer=issue-stream&stream_index=10

![Image](https://github.com/user-attachments/assets/601ebcd5-a7dd-4881-bcdc-e4518e3908e6)

https://sandbox.sentry.io/issues/6562584726/?project=4508160335675392&query=is%3Aunresolved&referrer=issue-stream&stream_index=9

![Image](https://github.com/user-attachments/assets/c8f215ee-b335-43d3-b1c1-11b317c0406f)

It's also not ideal that we have `(No error message)` but that's product thing, we need to figure out a proper way to add a title there at least for demo purposes but ideally improve this in the product.

PR #851, opened 2025-06-06
aspnetcore disable blocking detection

bruno-garcia:
I think it's best we disable until it gets a bit better:

See:
* https://github.com/sentry-demos/empower/issues/850
* https://github.com/getsentry/sentry-dotnet/issues/4262
* https://github.com/getsentry/sentry-dotnet/issues/4263


realkosty (approved):
thanks!

PR #852, opened 2025-06-10
Add logs to flask

lukemun:
- Migrate print to logs.info
- added sentryclirc to gitignore (need to change auth token for different empower projs now)

Tested in staging

image

image

### DIFF
--- a/flask/src/queues/tasks.py
+++ b/flask/src/queues/tasks.py
@@ -1,7 +1,6 @@
 from .email_subscribe import app
 import time, random
-import sentry_sdk
-
+import logging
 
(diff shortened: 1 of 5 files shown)


flask/src/queues/tasks.py
```diff
@@ -1,7 +1,6 @@
 from .email_subscribe import app
 import time, random
-import sentry_sdk
```
realkosty, line 3: are you sure we want this? don't we need the sdk for transaction auto-instrumentation

lukemun: it's unused here, in email_subscribe is where sdk is initialized

realkosty (approved):

PR #855, opened 2025-06-12
Fix sql vuln

lukemun:
- tested flow in staging


realkosty (approved):
thanks!

once injection detection is live in EA let's modify the second one to actually create an issue in Sentry. We can use something that's guaranteed to be safe (e.g. a constant, not parsed from query params)

PR #857, opened 2025-06-13
[flask, seer] autofixable flagship error - AttributeError - OPTION 1

realkosty:
This is designed not to break Critical Experiences (see `validate_inventory`)

This is mutually exclusive with Option 2 (https://github.com/sentry-demos/empower/pull/858) we have to pick one or the other depending on which one provides a better Seer demo.

This one can and should be chained with `raise ... from ...` just like Option 2. once we confirm that it doesn't interfere with Seer, of course.

# Testing
### standard_checkout_fail
http://localhost:3000/?cexp=standard_checkout_fail&se=attributeerror-cexp-standard_checkout_fail

Screenshot 2025-06-13 at 3 50 01 PM

[flagship error, N+1 query](https://kosty.sentry.io/issues/?project=4506005091319808&query=se%3Aattributeerror-cexp-standard_checkout_fail&referrer=issue-list&statsPeriod=30d)

###  checkout_success
http://localhost:3000/?cexp=checkout_success&se=attributeerror-cexp-checkout_success

Screenshot 2025-06-13 at 3 46 47 PM

[no errors, only N+1 query](https://kosty.sentry.io/issues/?project=4506005091319808&query=is%3Aunresolved%20se%3Aattributeerror-cexp-checkout_success&referrer=issue-list&statsPeriod=30d)

PR #858, opened 2025-06-13
[flask, seer] autofixable flagship error - UnboundLocalError - OPTION 2 [chained]

realkosty:
Implemented as chained exception (`raise ... from ...`). This way SEs are not confused and will see familiar-ish inventory error. Same can be done for Option 1 (https://github.com/sentry-demos/empower/pull/857)

This is designed not to break Critical Experiences (see `validate_inventory`)

This is mutually exclusive with Option 1 (https://github.com/sentry-demos/empower/pull/857) we have to pick one or the other depending on which one provides a better Seer demo.

Screenshot 2025-06-13 at 4 16 17 PM

# Testing
### standard_checkout_fail
http://localhost:3000/?cexp=standard_checkout_fail&se=unboundlocalerror2-cexp-standard_checkout_fail

Screenshot 2025-06-13 at 3 50 01 PM

[flagship error, N+1 query](https://kosty.sentry.io/issues/?project=4506005091319808&query=se%3Aunboundlocalerror2-cexp-standard_checkout_fail&referrer=issue-list&statsPeriod=30d)

###  checkout_success
http://localhost:3000/?cexp=checkout_success&se=unboundlocalerror2-cexp-checkout_success

Screenshot 2025-06-13 at 3 46 47 PM

[no errors, only N+1 query](https://kosty.sentry.io/issues/?project=4506005091319808&query=is%3Aunresolved%20se%3Aunboundlocalerror2-cexp-checkout_success&referrer=issue-list&statsPeriod=30d)

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -231,9 +231,9 @@ def checkout():
                         fulfilled_count += 1
                     else:
                         out_of_stock.push(f'Item #{i}')
-    except Exception:
+    except Exception as err:
         sentry_sdk.metrics.incr(key="checkout.failed")
-        raise
+        raise Exception("Error validating enough inventory for product") from err
 
     if len(out_of_stock) == 0:
         result = {'status': 'success'}
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -12,7 +12,7 @@ from flask_caching import Cache
 from statsig.statsig_user import StatsigUser
 from statsig import statsig, StatsigOptions, StatsigEnvironmentTier
 import dotenv
-from .db import get_products, get_products_join, get_inventory
+from .db import decrement_inventory, get_products, get_products_join, get_inventory
 from .utils import parseHeaders, get_iterator, evaluate_statsig_flags
 from .queues.tasks import sendEmail
 import sentry_sdk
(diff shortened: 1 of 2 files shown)


flask/src/main.py
```diff
@@ -208,25 +208,42 @@ def checkout():
...
-        if len(inventory) == 0 or len(quantities) == 0:
-            raise Exception("Not enough inventory for product")
-
-    response = make_response("success")
-    return response
+    fulfilled_count = 0
+    out_of_stock = [] # list of items that are out of stock
+    try:
+        with sentry_sdk.start_span(op="process_order", description="function"):
+            if validate_inventory:
```
lukemun, line 222: is checkout_success whats used for cexp? since all the logic is nested under this `if` would validate_inventory=false result in an empty process_order span?

flask/src/main.py
```diff
@@ -208,25 +208,42 @@ def checkout():
...
+                quantities = cart['quantities']
+                inventoryDict = {x.productid: x for x in inventory}
+                for i, cartItem in enumerate(quantities):
+                    if cartItem in inventoryDict and inventoryDict[cartItem].count >= quantities[cartItem]:
+                        decrement_inventory(inventoryDict[cartItem].id, quantities[cartItem])
+                        fulfilled_count += 1
+                    else:
+                        out_of_stock.push(f'Item #{i}')
+    except Exception:
+        sentry_sdk.metrics.incr(key="checkout.failed")
```
lukemun, line 235: should we still be using the metrics product?

realkosty: oh good catch
out of scope here - let's clean up everywhere as a separate PR
we could replace with span attributes instead of removing but not sure how much that would add to our CritExp story.. @ndmanvar what do you think?

lukemun (approved):
LGTM, regarding the metrics, aidan found it was breaking the nextjs app during checkout, though looks like it's ok here

[metrics breaking checkout in nextjs](https://github.com/sentry-demos/empower/pull/868)

realkosty:
> LGTM, regarding the metrics, aidan found it was breaking the nextjs app during checkout, though looks like it's ok here
> 
> [metrics breaking checkout in nextjs](https://github.com/sentry-demos/empower/pull/868)

oh good to know. Let's put this at the top priority for `flask` demo then to make sure it doesn't break if those APIs are fully shut down at some point

PR #864, opened 2025-06-17
Set project id for flask deployment

lukemun:
- project_id isn't specified locally causing tunnel to not setup correctly
- staging/prod project_id isn't required as we are already in that project

Testing
- deployed to local and staging successfully


realkosty (changes requested):
could we move that to `deploy.sh`? where is PROJECT_ID coming from?

lukemun:
> could we move that to `deploy.sh`? where is PROJECT_ID coming from?

[here](https://github.com/sentry-demos/empower-config/pull/37)

in deploy.sh it will run for all env though? we only need this for local

realkosty (approved):
ah right

PR #937, opened 2025-07-16
Add seer tag when flask

landenai:
When the backend service called during the checkout flow is flask, set the seerDemo tag to true. This will enable team members to find example issues for the seer demo easily.

**Testing:**

- This tag hsould only be set when the flask service is called during the checkout flow

### DIFF
--- a/react/src/tests/Checkout.test.js
+++ b/react/src/tests/Checkout.test.js
@@ -48,74 +48,71 @@ describe('Checkout Component', () => {
     // expect(screen.getByPlaceholderText(/123 Main Street/i)).toBeInTheDocument();
   });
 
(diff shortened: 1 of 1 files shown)


react/src/tests/Checkout.test.js
```diff
@@ -44,6 +48,72 @@ describe('Checkout Component', () => {
...
+  //   );
+
+  //   // Trigger checkout by submitting the form
+  //   fireEvent.submit(screen.getByRole('button', { name: /Complete order/i }));
+
+  //   await waitFor(() => {
+  //     expect(Sentry.setTag).not.toHaveBeenCalledWith('seerDemo', true);
+  //   });
+  // });
+
```
maxkosty: ???

PR #943, opened 2025-07-17
[react] remove dedupe integration to get more errors re-attempting checkout

realkosty:
# Testing 
`./deploy.sh --env=local react`


realkosty:
self-approving to unblock TDA testing after hours

PR #958, opened 2025-07-19
[tda] switch to new host

maxkosty:
# Goal
hoping this will fix the stuck/deadlock situation we've been experiencing seemingly after mounting old disk on more performant VM type. 

# Implementation
everything installed from scratch (incl. python3.10)
this VM only has ssh access through IAP (https://github.com/sentry-demos/empower/pull/957 accommodates that)

# Testing
```
./deploy_tda.sh
gcloud compute config-ssh
gcloud compute ssh empower-tda-and-crons-2
ps fxww -u kosty --sort -etime o pid,user,command
```

### DIFF
--- a/tda/deploy_tda.sh
+++ b/tda/deploy_tda.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-HOST=empower-tda-and-crons
-HOST_ZONE=us-central1-a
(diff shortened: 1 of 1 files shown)


maxkosty:
Will change new instance (`empower-tda-and-crons-3`) name to match the old one instead. Zone is the same

PR #971, opened 2025-07-24
[flask,tda] elevate react flagship above flask flagship

realkosty:
# Goal

Currently `flask` flagship gets about 1% more events likely because some lost on the client side
This change aims to tip the balance just enough so that `500 - Internal Server Error` comes on top

Screenshot 2025-07-24 at 3 57 34 PM


realkosty:
After some thinking - we should instead make 1 more backend autofixable and include it in flagship React via grouping in addition to `backendType:flask`. That's a cleaner solution

PR #1006, opened 2025-08-27
feat(logging): enhance Flask logging with structured logging 

lukemun:
• Migrate from print statements to structured logger.info calls
• Add comprehensive logging to all Flask endpoints with request context
• Update Sentry configuration to enable logs with proper integration
• Remove deprecated LoggingIntegration in favor of enable_logs parameter
• Upgrade sentry-sdk from 2.29.1 to 2.35.1 for improved logging support

### DIFF
--- a/flask/src/main.py
+++ b/flask/src/main.py
@@ -275,11 +275,7 @@ def checkout():
                 logging.debug("> inventoryItem.count %s", inventoryItem['count'])
                 if (validate_inventory and (inventoryItem.count < quantities[cartItem] or quantities[cartItem] >= inventoryItem.count)):
                     sentry_sdk.metrics.incr(key="checkout.failed")
-                    logger.warning('Processing /checkout - insufficient inventory', extra={
-                        'product': cartItem,
-                        'requested_quantity': quantities[cartItem],
-                        'available_inventory': inventoryItem.get('count', 0)
-                    })
+                    logger.warning('Processing /checkout - insufficient inventory')
                     raise Exception('Not enough inventory for product')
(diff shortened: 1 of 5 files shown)


flask/src/main.py
```diff
@@ -313,102 +426,211 @@ def get_api_request(key, delay):
...
     return response
 
 
 # compressed assets
 @app.route('/compressed_assets/<path:path>')
 def send_report_configured_properly(path):
+    logger.info('Received /compressed_assets request', extra={
+        'endpoint': '/compressed_assets',
+        'file_path': path
+    })
```
maxkosty, line 605: This adds 45 multi-line logging statements. I personally don't love having to wade through all this while maintaining the app.

Q1: do we really need all of them? What's the demo flow we are catering for here? I think we may want some log line in most spans in `/products` in case somebody goes around randomly clicking in the waterfall. Then maybe another 3-5 in `/checkout`. Plus whatever you find useful to actually debug the application (e.g. caches), or anything that's "close" to some demo path, e.g. N+1 query.

Q2: can we make them less verbose? One way is [something like this](https://github.com/getsentry/sentry-helper-slack-bot/blob/547f8629173c4125ade262261adc364d3bf20039/app.js#L297-L299). In a real application this is a must IMO unless you want your code to be 50% instrumentation.

maxkosty (approved):
I'm not a fan of the amount of boilerplate but feel free to ship if you don't have time to improve this

maxkosty:
There seems to be a duplicate PR #1007 , can we close this one?

PR #1013, opened 2025-09-02
add cart total in failed and succesful checkout

serglom21:
move `checkout.order.total` attribute declaration up so that it gets attached to events when checkout succeeds and fails. Currently, only gets attached if it succeeds. But it would be useful to have it to build widgets for "Total Revenue Lost" which seems to be a popular use case

Screenshot 2025-09-02 at 5 21 35 PM


maxkosty (approved):
Great idea!

PR #1025, opened 2025-09-16
Laravel: update Empower Plant demo App to Laravel 12

constantinius:


maxkosty (changes requested):

maxkosty (approved):
Thank you! 
please follow up with deleting the old `laravel` and replacing it with `laravel-8-backup`

PR #994, opened 2025-08-20
[demo-refresh] Add demo refresh changes to vue

serglom21:
## Frontend Monitoring

- [x] Enable tracing, profiling, logs, session replay, and session health. Ensure all relevant auto-instrumentation is enabled by reviewing the SDK options.

- [x] Modify a flagship error line (e.g., change whitespace) to ensure a Suspect Commit is visible for the next 12 months.

- [x] Integrate Browser profiling using Sentry.browserProfilingIntegration().

## Backend Routing System
- [x] Implement dynamic backend type detection via URL parameters (e.g., ?backend=flask).

- [x] Add support for multiple backend services: Flask, Express, Spring Boot, ASP.NET Core, Laravel, Ruby, and Ruby on Rails.

- [x] Configure backend URLs using environment variables.

- [x] Provide global access to the backend configuration via window.BACKEND_URL.

## Experiment & Tag Tracking
- [x] Implement SE tag tracking

- [x] Randomize customer types (medium-plan, large-plan, small-plan, enterprise).

- [x] Implement CEXP experiment tracking with conditional logic.

- [x] Add frontend slowdown experiment controls.

- [x] Enable API type switching (e.g., products vs. products-join).

- [x] Set up rage click and user feedback experiments.

- [x] Track user emails with the Sentry user context.

## Infrastructure Configuration
- [x] Configure Vite dev server headers for development profiling.

- [x] Configure App Engine production headers for production profiling.

- [x] Implement environment variable validation and templates.

- [x] Create utility functions for component access to experiment flags.

PR #1425, opened 2026-04-10
add Vue service to DNS load balancer config

serglom21:
## Description

Adding Vue domain to load balancer config as part of Vue Demo Refresh (#994 )


maxkosty (changes requested):
please update `staging.env`, `production.env` and `.local.env.base` 
#994 only adds vars to `staging.env` 
make sure to follow the current pattern with <PROJECT>_SUBDOMAIN e.g. https://github.com/sentry-demos/empower/blob/991204bfe581fe7d104cf6013793f4fc9ee2fd64/production.env#L107-L112 (what Seer comment below is referring to).

otherwise looks good!

ndmanvar (approved):

PR #1461, opened 2026-05-07
fix: promo code apply margin

nw-sentry:
Button has a default margin of 8px
The "Apply" button has a margin top zero which makes the button and textbox to be mis-aligned vertically

Before
Screenshot 2026-05-06 at 7 37 52 PM

After
Screenshot 2026-05-06 at 7 38 08 PM


cstavitsky (approved):
LGTM 👍

PR #1515, opened 2026-06-09
Add demo-error query params + propagate se/user tags to agent spans

ndmanvar:
## Summary

Adds explicit, deterministic query-param control over the plant agent's demo errors (replacing the in-agent random injection), forwards operator-facing flags from the web app, and propagates request-context tags (`se`, `customerType`, `cexp`, `user.email`) onto every span in the agent trace.

### Demo error params

The agent fails specific plant lookups only when explicitly told — no random self-injection. The **backend** params are domain-named so the trigger doesn't stand out in Sentry span / attribute comparison; the **web app** exposes friendlier operator flags that map to them.

| Operator flag (page URL) | Forwarded to agent as | Behavior |
|--------------------------|-----------------------|----------|
| `?agent_advice_error=true` | `validate_plant_advice` | Hard **500** — `Could not get plant advice: File not found` |
| `?agent_info_error=true` | `validate_plant_info` | Soft tool error — `Could not get plant basic info: Unknown plant` |

Unset / anything else = false (no failure).

The ambient randomness for synthetic traffic now lives in the **TDA test**, which sets the flags at `PLANT_ADVICE_FAILURE_RATE = 0.2` and `PLANT_INFO_FAILURE_RATE = 0.1` — so real users never hit injected errors, only synthetic traffic does.

- `agent/app/utils.py` — two plain `bool` contextvars (`validate_plant_advice`, `validate_plant_info`); removed `maybe_throw` / random injection.
- `agent/app/api/routes.py` — parses both query params.
- `agent/app/tools/plant_base_info.py` — raises on the explicit flags instead of random throws.
- `react/src/components/ChatWidget.jsx` — maps `agent_advice_error` → `validate_plant_advice` and `agent_info_error` → `validate_plant_info` onto the `/buy-plants` request.
- `_tda/desktop_web/test_ai_agent.py` — drives the failure rates (0.2 / 0.1) via the operator flags.

### Propagate context tags onto gen_ai spans

`sentry_sdk.set_tag("se", ...)` only lands on the root `http.server` span, so the auto-instrumented `gen_ai.*` child spans didn't carry `se`. Added a `before_send_transaction` hook (`agent/main.py`) that mirrors `se` / `customerType` / `cexp` / `user.email` onto every span's `data` and the trace root.

## Testing

Verified locally (react :3000, agent :8093, flask :8083 via `./deploy --env=local`):
- `?agent_advice_error=true` → forces the 500 when the LLM routes through the plant-advice tool.
- Agent honors `validate_plant_advice` / `validate_plant_info` directly.
- `before_send_transaction` confirmed to copy the tags onto child `gen_ai.*` spans.

> Note: the agent only errors when the LLM (gpt-5-mini) actually routes through the relevant tool, so it isn't perfectly deterministic per request — the flags control how often the failure is attempted, and the TDA test exercises them across many iterations.

## Addressed review feedback
- Removed in-agent random error injection; randomness now lives in the TDA test (per review).
- Domain-named the backend params so the trigger doesn't break the 4th wall in attribute comparison.
- Simplified the contextvar to a plain `bool` (unset = false).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agent/app/tools/plant_base_info.py
+++ b/agent/app/tools/plant_base_info.py
@@ -5,7 +5,7 @@ from typing import Any
 
 from agents import FunctionTool
 
-from ..utils import agent_crash_mode, maybe_throw
+from ..utils import validate_plant_advice, validate_plant_info
 
 # Configure logging
 logging.basicConfig(level=logging.DEBUG)
@@ -38,7 +38,8 @@ def get_plant_basic_info(plant_names: list) -> str:
         },
     }
 
-    maybe_throw(0.1, Exception("Could not get plant basic info: Unknown plant"))
+    if validate_plant_info.get():
+        raise Exception("Could not get plant basic info: Unknown plant")
 
     # Collect info for each plant
     info_list = []
@@ -64,9 +65,8 @@ async def _invoke_plant_advice(context: Any, input_json: str) -> str:
     """Invoke the plant advice tool."""
     import json
 
-    if agent_crash_mode.get() is True:
+    if validate_plant_advice.get():
         raise Exception("Could not get plant advice: File not found")
-    maybe_throw(0.2, Exception("Could not get plant advice: File not found"))
 
     try:
         logging.debug(f"Invoking get_plant_basic_info with input: {input_json}")
--- a/agent/app/tools/plant_base_info.py
+++ b/agent/app/tools/plant_base_info.py
@@ -5,7 +5,7 @@ from typing import Any
 
 from agents import FunctionTool
 
-from ..utils import maybe_throw
+from ..utils import agent_crash_mode, maybe_throw
 
 # Configure logging
 logging.basicConfig(level=logging.DEBUG)
@@ -64,6 +64,8 @@ async def _invoke_plant_advice(context: Any, input_json: str) -> str:
     """Invoke the plant advice tool."""
     import json
 
+    if agent_crash_mode.get() is True:
+        raise Exception("Could not get plant advice: File not found")
     maybe_throw(0.2, Exception("Could not get plant advice: File not found"))
 
     try:
--- a/agent/app/utils.py
+++ b/agent/app/utils.py
@@ -1,6 +1,21 @@
+import contextvars
 import random
+from typing import Optional
+
+# Controls the demo error injection via the ?agent_crash query param. Tri-state:
+#   True  -> force the plant advice error 100% of the time
+#   False -> suppress all injected errors (never crash)
+#   None  -> no override; fall back to the default random probabilities
+# Set in the API route and read inside the tool, which is invoked by the agents
+# SDK; a contextvar propagates through the awaited async chain.
+agent_crash_mode: contextvars.ContextVar[Optional[bool]] = contextvars.ContextVar(
+    "agent_crash_mode", default=None
+)
 
 
 def maybe_throw(probability: float, exception: Exception) -> None:
+    # An explicit ?agent_crash=false disables injected errors entirely.
+    if agent_crash_mode.get() is False:
+        return
     if random.random() < probability:
         raise exception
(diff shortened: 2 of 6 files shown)


ndmanvar:
`agent_crash=false`
[link](https://team-se.sentry.io/explore/traces/trace/a067135f47394aeeb92c579ead198e72/?fov=0%2C84469.40014648438&node=span-8ddc4e0d9a909a24&query=se%3A%EF%80%8DContains%EF%80%8Dneil-local-11&source=traces&statsPeriod=10m&targetId=9b866c4a09d649c8&timestamp=1780963109)
image

ndmanvar:
`agent_crash=true`
[link](https://team-se.sentry.io/issues/trace/61ac8c2304cc461499c4a4e06d58b80b/?groupId=7537721124&node=span-9c7cbff5c7ebed85&pageEnd=2026-06-09T11%3A55%3A01.595&pageStart=2026-06-08T11%3A55%3A01.595&project=4511531739185152&query=is%3Aunresolved%20se%3A%EF%80%8DContains%EF%80%8Dneil-local&referrer=issue-stream&source=issue_details&timestamp=1780962901.501)
image

ndmanvar:
TODO: test w/ staging

agent/app/tools/plant_base_info.py
```diff
@@ -64,6 +64,8 @@ async def _invoke_plant_advice(context: Any, input_json: str) -> str:
     """Invoke the plant advice tool."""
     import json
 
+    if agent_crash_mode.get() is True:
+        raise Exception("Could not get plant advice: File not found")
     maybe_throw(0.2, Exception("Could not get plant advice: File not found"))
```
maxkosty: modify _tda/desktop_web/test_ai_agent.py to pass `agent_crash=true` with probability 0.2 and remove this line?

agent/app/utils.py
```diff
@@ -1,6 +1,21 @@
+import contextvars
 import random
+from typing import Optional
+
+# Controls the demo error injection via the ?agent_crash query param. Tri-state:
+#   True  -> force the plant advice error 100% of the time
+#   False -> suppress all injected errors (never crash)
+#   None  -> no override; fall back to the default random probabilities
```
maxkosty: unnecessary complexity, not set should mean false.

agent/app/tools/plant_base_info.py
maxkosty: the way it is right now an explicit agent_crash=false may still result in error with 20% chance

agent/main.py
```diff
@@ -14,6 +14,38 @@
 from app.api.routes import router
 from config import settings
 
+
+def propagate_context_to_spans(event, hint):
+    """Copy request-context tags onto every span's attributes.
```
maxkosty, line 19: curious what we use this for

minor concern - `agent_crash` tag this may show up in attribute comparison in the issue and break the 4th wall in a demo. Unless that's how you intend it. But if so we should use some made up "domain tag" like the ones we already have, e.g. `validate_inventory`.

react/src/components/ChatWidget.jsx
```diff
@@ -280,9 +280,17 @@ const ChatWidget = () => {
           requestHeaders['x-conversation-id'] = conversationIdRef.current;
         }
 
+        // Forward the agent_crash query param from the page URL to the agent
+        // so the backend can force the plant advice error on demand.
+        let buyPlantsUrl = `${AGENT_URL}/api/v1/buy-plants`;
+        const agentCrash = new URLSearchParams(window.location.search).get('agent_crash');
```
maxkosty: as long as agent shows up only on the landing index page (which is the case right now) this works

maxkosty (changes requested):

maxkosty (approved):

react/src/components/ChatWidget.jsx
ndmanvar: yep, somethign we should be aware about if we add agent to other pages/routes/etc.

PR #1615, opened 2026-07-31
fix(agent): route plant agents through OpenRouter

HazAT:
## Summary
- point the plant agent OpenAI client at OpenRouter (`https://openrouter.ai/api/v1`)
- rename deploy secret `AGENT_OPENAI_API_KEY` → `AGENT_OPENROUTER_API_KEY`
- switch models to OpenRouter ids (`openai/gpt-5-mini`, `openai/gpt-5-nano`)
- replace OpenAI-only `HostedMCPTool` with client-side `MCPServerStreamableHttp`
- add `mcp` dep; disable OpenAI tracing; set `store=False` for OpenRouter Responses

## Deploy notes
1. create GCP secret `AGENT_OPENROUTER_API_KEY` with the OpenRouter key
2. deploy `agent`

## Test plan
- [ ] deploy agent staging with OpenRouter secret
- [ ] hit `/api/v1/buy-plants` and confirm recommendations + MCP product lookup
- [ ] confirm no calls to `api.openai.com`


maxkosty:
tested with a [test key](https://openrouter.ai/workspaces/default/keys)
`./deploy --env=local react agent`

Getting this error: 
https://kosty.sentry.io/issues/7644719675/?project=4510718200774656&query=is%3Aunresolved&referrer=issue-stream

I forgot if error without a query parameter is expected or not

Screenshot 2026-07-31 at 9 16 04 AM

@ndmanvar could you please help take it from here while I'm away? Daniel already spent a lot of time just trying to get all the permissions to our github org/repo.

@HazAT will provide the AGENT_OPENROUTER_API_KEY for production.

HazAT:
@maxkosty no rush, let's deal with this when you are back - I keep the old keys alive for now

You can also then just create your own key

constantinius:
@maxkosty the Sentry issue you have states it uses `sentry-sdk==2.44.0` (which is ancient), thus the `sentry_sdk.ai.set_conversation_id()` is not available. The requirements.txt however uses `sentry-sdk[fastapi]==2.56.0` which should be exactly what we need. So I guess you have an outdated local setup?

constantinius:
Added `set_default_openai_api("chat_completions")` to the config — the Agents SDK was defaulting to OpenAI's Responses API and 404ing on `/responses`, since OpenRouter only implements Chat Completions. Works now. (92972f02)

